PR #3, opened 2015-06-12
Featureparity

davidagold:
This PR:
1. Adds analogous methods for the most of the primitives found in DataArrays/src/dataarray.jl
2. Ports a number of tests from DataArrays/test/dataarray.jl, though does not contain complete coverage
3. Adds an inner constructor that checks that the lengths of the `isnull` and `values` fields are equal
4. Adds outer constructors analogous to those of DataArrays, plus a new outer constructor whose purpose is similar to that of the `@data` macro.

I promise all future PRs will only involve self-contained revisions/additions... >_>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,3 @@
+language: julia
+julia:
+  - nightly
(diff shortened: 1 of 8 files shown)


johnmyleswhite:
:tada:

davidagold:
Haha, indeed!

PR #4, opened 2015-06-13
Remove levels (TBF in separate package)

davidagold:
This PR removes the 'levels' method from src/03_primitives.jl. It also adds a little note about the 'reverse' method, which `NullableArray` currently inherits from `AbstractArray` in an attempt to keep track of inheritance and also record notes about possibilities for performance improvements with specialized methods.

### DIFF
--- a/src/03_primitives.jl
+++ b/src/03_primitives.jl
@@ -253,7 +253,7 @@ end
 # Use ready-made method for AbstractArrays or implement method specific to
 # NullableArrays, possibly for performance purposes?
 
-# ----- levels ---------------------------------------------------------------#
-function levels(a::AbstractArray) # -> Vector{T}
-    return unique(a)
-end
+# ----- Base.reverse ---------------------------------------------------------#
+
+# Use ready-made method for AbstractArrays or implement method specific to
+# NullableArrays, possibly for performance purposes?


johnmyleswhite:
We should benchmark implementations of `reverse`. I imagine we can improve over the `AbstractArray` version, but I doubt the improvement is more than 5x.

davidagold:
Actually, the difference turns out to be quite substantial, at least for the following implementation and test case:

``` julia
using NullableArrays
srand(1)

X = NullableArray(rand(5_000_000), rand(Bool, 5_000_000))

function reverse1(X::NullableArray)
    return(NullableArray(reverse(X.values), reverse(X.isnull)))
end

Base.reverse(X)
reverse1(X)

@time(Base.reverse(X))
@time(reverse1(X))
isequal(Base.reverse(X), reverse1(X))
```

yields

```
julia> include("/Users/David/.julia/v0.4/NullableArrays/perf/reverse.jl")
 280.878 milliseconds (19999 k allocations: 501 MB, 23.95% gc time)
  17.877 milliseconds (9 allocations: 43946 KB, 6.41% gc time)
true
```

johnmyleswhite:
Is that with `@inline` turned on for both `getindex` and `setindex`?

davidagold:
Ah, no, you're right. That actually brings things precisely in line with your estimate:

```
julia> include("/Users/David/.julia/v0.4/NullableArrays/perf/reverse.jl")
 100.103 milliseconds (13 allocations: 43946 KB, 32.84% gc time)
  20.808 milliseconds (9 allocations: 43946 KB, 17.80% gc time)
true
```

davidagold:
Is it inlining per se that causes the performance jump, or are there also other optimizations that the compiler can perform when it sees that indexing operations are inlined? I don't understand why so much memory is allocated when inlining is disabled. Does the lower level code allocate temporary space to hold the result from each call to `getindex`?

johnmyleswhite:
Nice.

I think the strategy of benchmarking every function relative to the implementation for `Array` is going to be a great way for us to know when we're near the performance limits we can reach and when we're much too far away. I suspect type-inference + boxing issues will account for almost all the places when we see >20x slowdowns.

davidagold:
Agreed.

Or, now that you mention boxing, is it that the program has to box the result of `getindex`?

johnmyleswhite:
That's a good question. I don't know offhand why inlining would change the amount of memory allocation.

johnmyleswhite:
It really shouldn't have to box the result of `getindex` unless we've made a mistake (or failed to understand how the issue 10525 changes work in Julia).

johnmyleswhite:
In general, there should be boxing any place that Julia thinks a computation needs to check type tags at run-time.

davidagold:
That's what I thought. Well, I'd like to review the 10525 changes in depth anyway, so maybe I will come across something there.

In the meantime, I imagine that inlining should be turned on for the indexing operations in src/04_indexing?

davidagold:
I feel like understanding where all that memory allocation is coming from is precisely one of those performance issues worth including in a discussion of the path to NullableArrays. Not to mention it'll make development easier, since I imagine a similar issue may well arise in attempts to implement other features.

johnmyleswhite:
Yeah, we should adopt the following guidelines:
- Benchmark the perf wins from inlining any deep core functions like `getindex` and `setindex!`. Those two definitely need to be inlined. Whenever we see huge wins, we should probably inline other core functions. (I doubt and hope there will be many more.)
- Avoid returning `Nullable{Union()}` anywhere that returns `Nullable{T}` for some concrete type `T`
- Continue to use `Array{Bool}` for the `isnull` mask. For some clever algorithms, `BitArray` could give a 2x speed boost, but it's also a 10x loss for less clever algorithms. We should probably follow minimax logic and try to select implementations that minimize the performance of the worst performing reasonable algorithm (within limits).

johnmyleswhite:
I agree. I'm really glad we figured a lot of this out and documented it early on. The indexing issues really hit at the heart of the problem: 
- Our decisions on inlining mostly affect downstream code.
- Different implementations of the same basic algorithm can have wildly different performance and we need to try to make all of them workable.

davidagold:
Indeed! All the above sounds good. I think I am going to try to emulate the benchmarking framework you use in your DesignSpace repo as something of a standard for performance case studies in NullableArrays. Down the line it'd be really nifty to be able to generate visuals automatically for performance tests.

johnmyleswhite:
I'm working on improving our benchmarking tools already; we should hopefully be able to use codespeed  to get automatic visualizations.

davidagold:
Awesome. Histograms for multiple runs? Kind of icing on the cake, but would also be nifty.

johnmyleswhite:
The goal is visualizations on par with http://www.serpentine.com/criterion/

davidagold:
Cool!

PR #7, opened 2015-06-15
Indexing

davidagold:
This PR:

-adds `Base._checkbounds` and `Base.to_index` methods to better take advantage of the `AbstractArray` indexing regime 
-includes a `nullify` method to set a null index for a `NullableArray` 
-includes two unsafe indexing methods for `NullableArray`s
-includes indexing-related tests ported from the `DataArrays` testing suite.

### DIFF
--- a/src/04_indexing.jl
+++ b/src/04_indexing.jl
@@ -1,6 +1,8 @@
 # NullableArray is dense and allows fast linear indexing.
 import Base: LinearFast
 
+#----- GENERAL INDEXING METHODS ----------------------------------------------#
+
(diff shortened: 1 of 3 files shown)


johnmyleswhite:
Is `nullify` redundant with passing `setindex!` a `Nullable` argument?

davidagold:
It is -- I forgot about that behavior. Come to think of it, that explains why `Base.reverse` can return a proper `NullableArray` in the presence of null indices.

Although, I kind of like `nullify` as a shorthand. But then probably it would be better in this capacity implemented as something like

``` julia
nullify!(X, I::Int...) = setindex!(X, Nullable{eltype(X)}(), I...)
```

PR #21, opened 2015-06-26
Add `@^` macro, respective tests

davidagold:
This PR introduces the lifting macro `@^`, which can be used to annotate function calls as so: `@^ f(x) Int`. The latter type argument designates what type of `Nullable{T}` ought to be returned in case one of the function call arguments is null. If the type argument is omitted, then `Nullable{Union{}}` will be returned in the case of a null argument.

``` julia
julia> macroexpand(:( @^ f(x, y) Int ))
:(if y.isnull || x.isnull # line 27:
        Nullable{Int}()
    else  # line 29:
        Nullable(f(x.value,y.value))
    end)

julia> macroexpand(:( @^ f(x, X[i]) + g(y, Y[i]) ))
:(if ((y.isnull || X[i].isnull) || x.isnull) || Y[i].isnull # line 27:
        Nullable{Union{}}()
    else  # line 29:
        Nullable(f(x.value,X[i].value) + g(y.value,Y[i].value))
    end)
```

### DIFF
--- a/src/NullableArrays.jl
+++ b/src/NullableArrays.jl
@@ -6,6 +6,9 @@ module NullableArrays
            NullableVector,
            NullableMatrix,
 
+           # Macros
+           @^,
+
            # Methods
            dropnull,
            anynull,
@@ -22,4 +25,5 @@ module NullableArrays
     include("04_indexing.jl")
     include("05_map.jl")
     include("nullablevector.jl")
+    include("lift.jl")
 end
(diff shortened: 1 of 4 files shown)


johnmyleswhite:
Given that naive users won't understand the dangers of `Nullable{Union{}}`, I'm a little worried that the untyped form of lift is a performance trap.

davidagold:
Do you think it'd be better just to require that the user supply a fallback type? I'd be down with that. 

Another alternative is to try doing something with `Base.return_types`, but this may get complicated for nested function calls.

johnmyleswhite:
I think requiring a fallback type is the easiest way.

PR #26, opened 2015-06-29
Revise lift macro `@^`, respective tests

davidagold:
This PR introduces the following revision to the lift macro `@^`:

-Require type argument when calling `@^`, remove default behavior to return `Nullable{Union{}}` when type argument is omitted.

This revision is in response to suggestions made in https://github.com/johnmyleswhite/NullableArrays.jl/pull/21.

### DIFF
--- a/src/lift.jl
+++ b/src/lift.jl
@@ -7,9 +7,7 @@ macro ^(call, T...)
         throw(ArgumentError("@^: argument must be a function call"))
     end
 
-    if length(T) == 0
-        e_type = :(Union{})
-    elseif length(T) == 1
+    if length(T) == 1
         e_type = T[1]
     else
         throw(ArgumentError("@^: wrong number of arguments"))
--- a/test/lift.jl
+++ b/test/lift.jl
@@ -10,10 +10,10 @@ module TestLift
 
     @test isequal(@^(f(x), Int), Nullable(25))
     @test isequal(@^(f(y), Int), Nullable{Int}())
-    @test isequal(@^(f(y)), Nullable{Union{}}())
+    @test_throws ArgumentError eval(macroexpand(:( @^f(y) )))
+    @test_throws ArgumentError eval(macroexpand(:( @^ if x > 0; f(x) end )))
 
     @test isequal(@^(f(X[1]) + g(x, X[1]), Int), Nullable(11))
     @test isequal(@^(f(X[1]) + g(y, X[1]), Int), Nullable{Int}())
-    @test isequal(@^(f(X[1]) + g(y, X[1])), Nullable{Union{}}())
 
 end


johnmyleswhite:
What does this do when given an `if` clause?

davidagold:
Currently it throws an `ArgumentError` if fed an expression with head other than `:call`. However, if I disable that and tweak some things, I can currently get it to do this:

``` julia
julia> macroexpand(:( @^ if x > 4; f(x) end Float64 ))
quote  # /Users/David/.julia/v0.4/NullableArrays/src/lift.jl, line 24:
    if x.isnull # line 25:
        Nullable{Float64}()
    else  # line 27:
        Nullable(if x.value > 4 # line 1:
                f(x.value)
            end)
    end
end
```

I'm working on devising a sufficiently general means of getting that latter `Nullable` construction to live inside the target `if` branch's body. Other than that, does the rest of the behavior seem like what you would want? That is, ought the macro to treat all variables as `Nullable`s regardless of whether they're in the `if` conditions or branch bodies?

davidagold:
Success, I think:

``` julia
julia> macroexpand(quote
           @^ begin 
               if x > 4
                   f(x) + g(y)
               elseif x > 2
                   f(x) / g(y)
               else
                   f(x)*g(y)*h(z) 
               end 
           end Float64
       end)
quote  # none, line 2:
    begin  # /Users/David/.julia/v0.4/NullableArrays/src/lift.jl, line 24:
        if (y.isnull || z.isnull) || x.isnull # line 25:
            Nullable{Float64}()
        else  # line 27:
            begin  # none, line 3:
                if x.value > 4 # line 4:
                    Nullable(f(x.value) + g(y.value))
                else  # line 5:
                    if x.value > 2 # line 6:
                        Nullable(f(x.value) / g(y.value))
                    else  # line 8:
                        Nullable(f(x.value) * g(y.value) * h(z.value))
                    end
                end
            end
        end
    end
end
```

We'll want to make explicit in the documentation that this lifting scheme is premised on the assumption that lifted functions do not take non-`Nullable` arguments to `Nullable` arguments, since otherwise one would require lifting within lifting. But I think that assumption fits well with the definition of "lift" that we've been considering.

johnmyleswhite:
Great.

I think we may want to be very careful about shipping this macro since it might have edge cases whose correctness we can't guarantee. We should definitely note that this is an experimental feature.

johnmyleswhite:
On that note, we actually might want to pull the lifting stuff into a separate package so that we can easily deprecate it if it turns out to be brittle. That's what we did with DataFramesMeta, which worked out really well.

davidagold:
I'd be down with a NullableMeta package. EDIT: Actually, if it's not taken, Lifting.jl or Lift.jl both have nice rings to them.

It definitely is hard to predict what sort of crazy expressions folks will want to lift. For that reason I was inclined to restrict its application solely to function calls, but lifting entire blocks where applicable is definitely convenient and helps to offset the additional keystrokes from specifying a fallback type parameter.

I suppose I may as well add support for other blocks?

johnmyleswhite:
Sounds like a good solution to me.

The problem that you need to write custom code for every kind of expression is exactly what makes this stuff so wacky. (There's also another problem, which is that you're lifting needs to not lift any function that's already lifted -- which can't be done with macros.)

Trying to add support for other stuff would be great.

johnmyleswhite:
Here's my current thinking on automatic lifted operators for infix operations: https://gist.github.com/johnmyleswhite/8ace4451357a9764ecfa

davidagold:
That looks in line with what I was thinking. Are you going to try to merge that into Base?

Two questions:
1. Why aren't you using the `||` operator for these?
2. What will happen for isbits-false types? From my experimentation with `@^`, isbits-false types seem to require branches to avoid accessing undefined referents. Are you looking into other ways to handle them?

johnmyleswhite:
`||` isn't an operator, but a magic piece of Julia's language semantics. (It's not even a function since it doesn't evaluate all arguments eagerly.)

I don't know how to make isbits-false types fast since you need to know the return type of applying the function to the values, but you can't do that.

davidagold:
Ahh. Well, is there a reason to prefer `|` in any case?

Right. So, until functions are overhauled, the best we could do is require that users specify an explicit fallback type, as in the case of `@^`? But it's not clear that's even worth implementing at this point?

johnmyleswhite:
`|` is what we should use for expressing vectorized boolean operations.

Yeah, my sense is that lifting is going to be pretty brittle until function types get changed a bit. Still worth experimenting with. One thing we should do is write out some workflows and see where lifting is really needed. I kind of suspect it's most helpful in SQL-style expressions, but we can use SQLite for that without needing any engineering resources to get things right.

PR #27, opened 2015-06-29
Experiment with presentation of `Nullable`s

davidagold:
This PR introduces an experimental means of displaying both empty and present `Nullable` values. The goal is to find some means of presenting such values so that the world `Nullable` does not clutter the screen when printing and showing `NullableArray`s. The scheme introduced here is that a present `Nullable{T}` value is shown as the interpolated value field followed by a question mark:

``` julia
julia> show(Nullable(5))
5?
```

and a missing `Nullable{T}` value is shown as `?{T}`:

``` julia
julia> show(Nullable{Int}())
?{Int64}
```

I know this breaks with the C# notation and such, but I don't like using the same pattern to represent present and null values, in part because of the ambiguity when one wants to show `Nullable(Int)`, although I suppose the meaning can be inferred easily enough from context. I'm not wedded to the current design by any means. But I do want to start somewhere.

### DIFF
--- a/src/NullableArrays.jl
+++ b/src/NullableArrays.jl
@@ -25,5 +25,6 @@ module NullableArrays
     include("04_indexing.jl")
     include("05_map.jl")
     include("nullablevector.jl")
+    include("io.jl")
     include("lift.jl")
 end
--- a/src/io.jl
+++ b/src/io.jl
@@ -1,0 +1,6 @@
+# Experiments with different means of printing present and missing Nullable
+# values to the screen
+
+function Base.show{T}(io::IO, x::Nullable{T})
+    print(io, ifelse(x.isnull, "?{$T}", "$(x.value)?"))
+end


quinnj:
I was meaning to talk to you guys about this. Ultimately, it would be great to also have a Nullable literal syntax where you could write the same output at the REPL to get a Nullable object, i.e.

``` julia
julia> typeof(5?)
Nullable{Int64}

julia> typeof(Int64?)
Nullable{Int64}
```

now, obviously `Int64?` wouldn't work to get a `Nullable{Int64}()` value, because it actually means `Nullable{DataType}(Int64)`. A few other ideas:
- `?::Int64` (utilizing the `::` syntax that is very "typey"
- `{Int64}?` similar to @davidagold's syntax above
- `?{Int64}?`

nalimilan:
@davidagold I don't think there's a possible confusion between `Nullable{Int}()` and `Nullable(Int)` thanks to the curly braces. So I'd say you could show it as `{Int64}?`, for consistency with non-null values.

That could indeed be a nice syntax for writing this object too, though the gain from `Nullable{Int}()` to `{Int}?` is certainly less than from `Nullable(5)` to `5?`.

davidagold:
@nalimilan You're right. I'm not really sure what I was thinking. I'm all for showing `Nullable{Int}()` as `{Int}?`.

@quinnj I agree that a shorthand for producing `Nullable` literals is desirable. It's not clear what will be best, though. And things get difficult because of `?`'s role as the ternary operator.

johnmyleswhite:
I'm really glad we're experimenting with this idea, but I think we should probably leave the definition of this `show` method to Base. Because this approach involves monkey-patching Base's definition, loading NullableArrays raises warnings.

quinnj:
Though I'm still a big fan of Mike Nolta's [suggestion](https://github.com/JuliaLang/julia/pull/9356#issuecomment-67085790)

``` julia
julia> t = Nullable(1)
¯\_(ツ)_/¯(1)
```

But seriously, I read back over that issue and there was some hesitation to stealing the `?` and worry that this is too new a feature. Do you think that still holds, John? Should we make another push for `?` or come up with some other ideas?

johnmyleswhite:
I think making a push for `?` might be more successful this time around, but it's still a risky proposition.

That said, I think any solutions to this problem have to happen in Base in the parser.

My instinct is to leave this problem for the very last step. When we've got awesome stuff and the only remaining hurdle is syntax, we'll have a stronger case for changing the parser.

nalimilan:
I agree that changing the parser will only make sense when everything is in place, so that we can make a convincing case that a deeper integration in the syntax is deserved.

quinnj:
Great thoughts guys; definitely not something to rush.

PR #28, opened 2015-06-29
Extend `svd`, `eig` methods for NullableMatrix

davidagold:
This PR ports extensions of `Base.svd` and `Base.eig` from https://github.com/JuliaStats/DataArrays.jl/blob/master/src/linalg.jl. The differences include:

-The methods defined for argument `M::NullableMatrix` fail by default if `anynull(M) = true`; the user must include a keyword argument `impute = true`.
-`findnull` and `global_mean` are implemented as generated functions that take `NullableArray` arguments of arbitrary dimensions.

Also adds dependency methods:
in `src/03_primitives.jl`
-`findnull`

in `src/linearalgebra.jl`
-`impute!`
-`global_mean`
-`nullsafe_rowmeans`

Also include tests.

### DIFF
--- a/src/03_primitives.jl
+++ b/src/03_primitives.jl
@@ -130,6 +130,27 @@ end
 
 allnull(X::NullableArray) = all(X.isnull) # -> Bool
 
+# ----- findnull -------------------------------------------------------------#
+
+@generated function findnull{T, N}(X::NullableArray{T, N})
+    if N == 1
+        e_res = :(res = Int[])
+        e_push! = :(push!(res, i_1))
+    else
+        e_res = :(res = NTuple{$N, Int}[])
+        e_push! = :(push!(res, @ntuple $N i))
+    end
+    return quote
+        $e_res
+        @nloops $N i X begin
+            if (@nref $N X i).isnull
+                $e_push!
+            end
+        end
+        return res
+    end
+end
+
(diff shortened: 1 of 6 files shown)


johnmyleswhite:
We  should probably remove the matrix imputation stuff from `NullableArrays`. It's too contentious and we don't want people to rely upon it.

davidagold:
Sure. What do you think we can keep? `findnull` seems like it'd be a good utility. Do we simply replace the `svd`, `eig` methods with something like

``` julia
function svd(X::NullableMatrix)
    if anynull(X)
        throw(NullException())
    else
        svd(X.values)
    end
end
```

Speaking of `NullException`s, the type currently has no `msg` field, so they can't be thrown with messages. Is this simply a matter of adding the field to the type definition in Base, or is it more involved?

johnmyleswhite:
Let me look through, but I would keep none of it. If people want complicated things like `svd`, they need to convert to a dense matrix or use an official matrix completion technique.

johnmyleswhite:
I wouldn't add a `msg` field to errors since they pulls them on the heap and makes functions that throw errors less amenable to inlining.

davidagold:
Got it. That's interesting, and kind of unfortunate given the already uninformative state of error handling. Oh well!

PR #41, opened 2015-07-16
Remove `src/map.jl`

davidagold:
As long as we aren't trying to implement any performance tricks (such as those employed in `base/broadcast.jl`), the present `src/map.jl` is entirely redundant with `base/map.jl` and oughtn't to be in the `NullableArrays` repo.

EDIT: I should also mention that the current implementation doesn't even succeed in implementing the `broadcast` trick, since it doesn't `eval` the internal `func` declarations. See https://groups.google.com/forum/?hl=en#!topic/julia-users/FkUxq0WNFo8 for me trying to understand why this makes a difference.


johnmyleswhite:
One question I have is whether we should provide a lifted version of map by default. It probably shouldn't be called map, but that might the version we should maintain above and beyond Base's implementation.

davidagold:
That's a good point. My thoughts on lifting have been shifting from "let's see how far we can empower users to lift functions at call sites via macros or internal machinery" to "let's either tell users to write lifted methods for their functions (and maybe provide a macro to help do so since that's easy enough), or let's just try make it so that the user never has to lift anything herself". A `maplift` function would fit the latter bill nicely.

johnmyleswhite:
I'd call it `mapnulls` to indicate that it's only one form of lifting.

davidagold:
In my draft section on lifting in the docs (on the `davidagold-readme` branch) I try to express that any means of systematically handling function calls on `Nullable` arguments in a way that yields conventional propagation semantics is "lifting". Are you concerned a user may mistake `maplift` it for some "canonical" means of lifting? 

I'm not sure that would be the worst thing in the world. Actually, I see a case to be made for going in the opposite direction, i.e. defining a generic `lift` method and have `lift(f, T, x::Nullable)` act like `@^ f(x) T` and having `lift(f, T, X::NullableArray)` act like `map(x -> @^ f(x) T, X)`, or, expressed more systematically, like `map(x -> lift(f, T, x), X)`. The implementation would be more or less exactly like that of `broadcast!` for performance, including caching the internal lifted methods. This approach seems very close to that suggested in https://github.com/JuliaLang/julia/pull/9446, except `map` is now `lift` [EDIT: Actually, after some more re-reading, I don't think it's "very" close, but still related]. `lift(f, T)` could return a generic function whose methods are all of `f`'s methods, but lifted, and `lift{N}(f, T, argtypes::NTuple{N, DataType})` could return just the method `f`takes on arguments of signature `argtypes`, but lifted. We could also provide mutating versions of the latter two methods to add the lifted methods to `f` itself.

The `@^` macro would still be the preferable lifting method over nested function calls, since `@^ f(g(h(x))) T` is easier than `lift(f, T3, lift(g, T2, lift(h, T1, x)))`, though the latter is necessary if `T1`, `T2`, `T3` are not pairwise identical. 

Anyway, I don't think `mapnulls` properly conveys the relevant functionality. Pretty much everywhere else "null(s)" indicates a concern primarily for null values, e.g. 'anynull', 'isnull', 'skipnull', etc. In that light, `mapnull` seems like you're either mapping null-ness over the Array or you're only mapping to the null values. I thought of `maplift` as an analogue to `mapreduce` -- first you map, then you lift. Or do you lift first then map? I guess it depends on whether we're mapping `lift(f, T)` over `X` or mapping `f` over `X` and lifting each function call individually.

However, one possible source of confusion from a generic `lift` is that one might call `lift(f, T, X::NullableArray)` where the intended method  of `f` acts on `Array` arguments and the user wants to  return, say, `Nullable(f(X.values))` if no null entries and `Nullable{T}()` if any null entries. For instance (and assuming that `sum` didn't already have a method for `NullableArray` arguments), one might intuitively expect for `X::NullableArray` that

```
lift(sum, T, X)
```

would act like the current implementation of `sum(X)`, but under the proposed semantics it would return `X`, at least if `T == eltype(X)`. Technically, such an expectation would not be in line with the conventional semantics for lifting we've adopted, i.e. if you have a method that acts on an `Array`, you lift it over a `Nullable{Array}`, not a `NullableArray` (I would not like to explain that verbally). But then again, neither is making `lift(f, T, X::NullableArray)` equivalent to `map(x -> lift(f, T, x), X)`. If this source of confusion outweighed the niceness of having one generic `lift` function, my vote would go towards `maplift` or `liftmap` or something that conveys lifting better than does `nulls`.

davidagold:
In any case, if the function won't be called `map`, do you think it's fine to go ahead with this PR for now?

johnmyleswhite:
The problem with using the word `lift` is that there are many reasonable senses in which the word `lift` could be used and we can't reserve one for `Nullable` without taking it away from everyone else. I worry that using the word `lift` without something like `liftnulls` is essentially the kind of punning that Julia discourages.

StefanKarpinski:
Would something like `lift(Nullable, ...)` be helpful? Another place where "lifting" occurs is in [Reactive](http://julialang.org/Reactive.jl/), and one could imagine writing `lift(Signal, ...)` there, which would potentially allow both to use the same verb. It would be good to have a decent general sense of what lifting generically means before doing that, but I suspect we may have enough examples at this point.

johnmyleswhite:
I think I'd argue that vectorization is also a form of lifting.

StefanKarpinski:
Agree.

davidagold:
@StefanKarpinski thanks for joining!

If I had to take a stab at it, I'd say that lifting an object from a domain A to a domain B entails generating a mapping from the semantics of the object in domain A to a semantics of the object in domain B, where this mapping could reasonably said to be "isomorphic" modulo whatever additional semantics that must be supported for engaging the object in domain B (for instance, branching on whether or not a `Nullable` argument supplied to a lifted operator is null).

If lifting a function over a `Nullable` argument, lifting a scalar operation over a `Vector` argument, and lifting a `Signal{T}` to a `Signal{U}` can all be sensibly fit into the above archetype, then I wouldn't find their common use of `lift` to be punning (I mean, if we used `lift` for vectorized operations, which I'm not recommending) any more than using `*` for string concatenation is punning. There's a fairly solid analogy between string concatenation and a general multiplicative binary operator, i.e. they're both generally non-commutative. How solid does the analogy between the different applications of `lift` need to be before it's no longer punning?

Also, as a practical matter, I can imagine at least two possible outcomes, at least as regards the verb `lift`:
1. Base or some other very basic package defines a `lift` verb around which can develop a use culture and philosophy against which other packages can evaluate any proposed extensions of `lift` for punning.
2. Packages define their own `lift` methods anyway, and there is nothing to delineate `lift` methods that more or less fit the archetype described above from `lift` methods that more or less pun amongst themselves.

Of course, if one doesn't think of the archetype above as actually an archetype, then 2 isn't really a problem over and above whatever punning occurs.

davidagold:
Also, my gut instinct (admittedly from the gut of somebody who is _not_ one of the language's founders, so take this FWIW) is that if `lift` in the cases of `Nullable` and `Signal` are sufficiently similar such that it would be acceptable to distinguish between them with `lift(Nullable, ...)` and `lift(Signal, ...)`, then it should be sufficient to distinguish between them with `lift(f, x::Nullable)` and `lift(s::Signal)`, i.e. just via plain ol' multiple dispatch. After all, isn't that the point of multiple dispatch?

StefanKarpinski:
Well, my (limited) understanding of the signal stuff in Reactive (maybe @shashi can correct me), is that you lift a function to get a signal, so you would have to provide the type that you want to get to distinguish.

davidagold:
Good point - we'd definitely want to make sure that there aren't any obvious conflicts. To that end, I'll point out that the current convention for lifting a function over `Nullable` arguments is that all arguments must be `Nullable`s, so a generic such use case would be `lift(f, xs::Nullable...)`, which seems like it doesn't conflict with [`Reactive.lift`](https://github.com/JuliaLang/Reactive.jl/blob/master/src/Reactive.jl#L362) since the former wouldn't (I don't think) steal any argument with a signature that could possibly be passed to `Reactive.consume`. But I'll definitely defer to Shashi on that particular issue.

davidagold:
Also, what about just having `lift` be a keyword argument (by default `false`) for `map(f, X::NullableArray)`?

johnmyleswhite:
That seems like the best approach to me.

shashi:
`Reactive.lift` has the signature `lift(f::Callable, s...)`.    The input args are converted to signals before calling `lift(f::Callable, s::Signal...)`. It is conceptually similar to `map` but I couldn't define a map method since map(f:: Callable, xs...) Is already taken in base, and I really wanted the automatic conversion to signal to be possible, and went with a different name.

Jeff said that the only thing he doesn't like about Escher is the word lift. So I made consume an alias. And it seems to be easier to explain to people in comparison to both lift and map. I think in your case lift could be a kwarg to map.

davidagold:
Thank you both for your feedback. I think kwarg to `map` is indeed the way to go.

davidagold:
@johnmyleswhite Presumably then we'd also want an analogous feature for `broadcast`?

johnmyleswhite:
I think making something analogous for `broadcast` is the right way to go.

davidagold:
Cool. I'm working with the `map` interface right now. I'd like to show that to you and get some feedback on it before doing the same for broadcast.

EDIT: "futz" isn't the right word.

johnmyleswhite:
Sure. If you e-mail me, I'll look at it tonight. Otherwise I could set up some time to chat by Skype tomorrow.

davidagold:
Great, thanks. I'll ping you when I introduce the PR, and if the strategy's
unclear we can talk about it via Skype some time soon.

On Mon, Jul 20, 2015 at 11:01 AM, John Myles White <notifications@github.com

PR #43, opened 2015-07-20
Maplift

davidagold:
The bulk of this PR involves overhauling the `map` interface for `NullableArray` arguments. The general strategy has been to implement the `@eval` trick used in `base/broadcast.jl` to improve performance while including the `lift` kwarg functionality in the generated internal function bodies. Right now I'm just putting the code out here for inspection; I'll include some performance comparisons later today. I want to point out a couple of things:
1. There is an asymmetric internal treatment of `map!` and `map` in that the former generates and caches the internal function based on on the value of the kwarg `lift`, whereas the latter does not. I made this design choice because `map` also calls `map_to!`, and both function bodies branch on the value of `lift`. It seemed to me that compiling different functions for both `map` and `map_to!` based on the value of `lift` was a bit much. On the other hand, it may be unnecessary to do so for `map!`, which should then be changed to receive the same treatment (as far as code generation and caching goes) as `map`.
2. There is an asymmetry between the location of the `::NullableArray` args in the signatures of `map!` ([here](https://github.com/johnmyleswhite/NullableArrays.jl/blob/d86b63caad871af9e2e8d1834093d8299a6849f9/src/map.jl#L144-L145)) and `map` ([here](https://github.com/johnmyleswhite/NullableArrays.jl/blob/d86b63caad871af9e2e8d1834093d8299a6849f9/src/map.jl#L159)). My thoughts were that we may as well make `map!`ing to a destination `NullableArray` fast for both `NullableArray` and non-`NullableArray` sources. However, this results in a parallel asymmetry between the "functionality" of the `lift` kwarg in `map!` and `map` when called on a signature including both `NullableArray` and non-`NullableArray` arguments: `map!(f, dest, A1::NullableArray, A2::Array)` will accept `lift` as a kwarg, but `map(f, A1::NullableArray, A2::Array)` will not since it falls back on the `map` defined in `base/abstractarray.jl`. The discrepancy isn't huge, since setting `lift=true` in the former will just result in an error. But maybe it would be conceptually more unified just to have our implementation of `map!` apply only to signatures of purely `NullableArrays` arguments (for the source arrays). Then `map!`ing from `NullableArray`s to a `NullableArray` will be fast but `map!`ing from arbitrary `AbstractArray`s to a `NullableArray` will be slow (since it will fall back on the `base/abstractarray.jl` implementation).
3.  Doing the code generation for `map` was a little tricky since `map` relies on `map_to!` and I generate code for both methods. ~~As a result, I end up passing the internally generated version of `map_to!` (`_map_to!` [in the code](https://github.com/johnmyleswhite/NullableArrays.jl/blob/d86b63caad871af9e2e8d1834093d8299a6849f9/src/map.jl#L163-L166)) around between the internally generated `map` function (`_map`) and itself for reasons described in the [comments](https://github.com/johnmyleswhite/NullableArrays.jl/blob/d86b63caad871af9e2e8d1834093d8299a6849f9/src/map.jl#L99-L100). This may make maintenance a bit more difficult. I feel like the same effect ought to be able to be achieved with a closure, but I can't figure out how...~~ I figured out [where to make the closure](https://github.com/johnmyleswhite/NullableArrays.jl/blob/46dac9927e74b8d1c22626e57dfe11e5c90155c8/src/map.jl#L120). Apart from eliminating the need for passing functions, this solution allowed me to eliminate the `gen_map_to!_function` method. Now the internal `_map_to!` method (which is named by a generated symbol that is hard-coded into itself and into its respective internal `_map` method and ) is generated by the same function as generates the internal `_map` method, which makes much more sense. 

EDIT (7/21): I've pushed my liftable version of `broadcast!` to this PR as well, since the functionality is more or less analogous and could be reviewed at the same time.


davidagold:
@johnmyleswhite using the profiling method in [this gist](https://gist.github.com/davidagold/0de522ac52973936ef9d) I get the following

```
julia> profile_map()
Method: map!(f, dest, src) (0 missing entries, lift=true)
  for Array{Float64}:           647.513 milliseconds (19999 k allocations: 305 MB, 20.21% gc time)
  for NullableArray{Float64}:    23.226 milliseconds (3 allocations: 112 bytes)

Method: map!(f, dest, src) (~half missing entries, lift=true)
  for NullableArray{Float64}:    37.107 milliseconds (3 allocations: 112 bytes)

Method: map(f, src) (0 missing entries, lift=true)
  for Array{Float64}:           426.723 milliseconds (10000 k allocations: 191 MB, 61.51% gc time)
  for NullableArray{Float64}:     1.662 seconds      (15000 k allocations: 391 MB, 68.63% gc time)

Method: map(f, src) (~half missing entries, lift=true)
  for NullableArray{Float64}:   207.942 milliseconds (2501 k allocations: 83030 KB, 34.60% gc time)
```

`map` is more computationally involved than `map!` because of the element type checking system employed, but I wonder if the awful performance in `map` demonstrated above is due to passing the `_map_to!` method around. That would make sense; in retrospect, it's kind of ironic that I used that solution to implement a workaround for functions passed as arguments being slow... I'll fiddle with that and see what I can do.

davidagold:
Well that's interesting. The current implementation is definitely broken, since if you take a look at the [gist](https://gist.github.com/davidagold/0de522ac52973936ef9d#file-gistfile1-jl-L33-L40) you can see that I don't actually set `lift=true` yet no error is thrown. So I have no idea what's going on there, but it may account for some of the slowness.

That issue appears to be fixed in my latest iteration on `src/map.jl`, in which I don't pass around any functions.

davidagold:
Things seem to be fixed. Here's the profiling of the latest iteration:

```
julia> profile_map()
Method: map!(f, dest, src) (0 missing entries, lift=true)
  for Array{Float64}:           677.053 milliseconds (19999 k allocations: 305 MB, 23.33% gc time)
  for NullableArray{Float64}:    13.570 milliseconds (3 allocations: 112 bytes)

Method: map!(f, dest, src) (~half missing entries, lift=true)
  for NullableArray{Float64}:    36.907 milliseconds (3 allocations: 112 bytes)

Method: map(f, src) (0 missing entries, lift=true)
  for Array{Float64}:           486.715 milliseconds (10000 k allocations: 191 MB, 67.98% gc time)
  for NullableArray{Float64}:   372.811 milliseconds (5000 k allocations: 119 MB, 85.43% gc time)

Method: map(f, src) (~half missing entries, lift=true)
  for NullableArray{Float64}:    95.149 milliseconds (2501 k allocations: 83030 KB, 26.93% gc time)
```

PR #50, opened 2015-08-12
Make `show`ing NullableArrays unobtrusive

davidagold:
This PR overrides a variety of `show` and `display` related methods in order to make the visual presentation of `NullableArray`s in the REPL less obtrusive. In particular, here's what we get:

``` julia
julia> X = NullableArray([1:10...], [true; fill(false, 9)])
10-element NullableArray{Int64,1}:
   NULL
  2    
  3    
  4    
  5    
  6    
  7    
  8    
  9    
 10    

julia> Any[X, X]
2-element Array{Any,1}:
 NullableArray{Int64,1}[NULL,2,3,4,5,6,7,8,9,10]
 NullableArray{Int64,1}[NULL,2,3,4,5,6,7,8,9,10]

julia> X = NullableArray(rand(2, 2), rand(Bool, 2, 2))
2x2 NullableArray{Float64,2}:
 0.00414697  NULL
  NULL       NULL
```

@johnmyleswhite @quinnj crossref: https://github.com/johnmyleswhite/NullableArrays.jl/issues/49

Probably a fair amount of the code introduced by this PR could be made unnecessary by instead overriding `show` for `Nullable` objects themselves, but others can make that call if material from this repo ever makes it into Base.

### DIFF
--- a/src/NullableArrays.jl
+++ b/src/NullableArrays.jl
@@ -29,4 +29,5 @@ module NullableArrays
     include("broadcast.jl")
     include("reduce.jl")
     include("statistics.jl")
+    include("show.jl")
 end
--- a/src/show.jl
+++ b/src/show.jl
@@ -1,0 +1,116 @@
+using Base: tty_size, alignment, print_matrix_row, strwidth, showcompact_lim
+
+type2string{T, N}(::Type{NullableArray{T, N}}) = "NullableArray{$T,$N}"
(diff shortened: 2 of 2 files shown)


quinnj:
Looking great! +1

nalimilan:
Great change indeed!

davidagold:
All right! Merging is go.

PR #52, opened 2015-08-17
importall Base.Operators

quinnj:


johnmyleswhite:
FWIW, I'd be more in favor of explicitly qualifying all operators.

quinnj:
That's cool. I think it was mainly `>=`, `<=`, `==` that I was seeing, so this was my quick fix. Probably easier for you or David to just push the fix.

davidagold:
Done https://github.com/johnmyleswhite/NullableArrays.jl/pull/54

quinnj:
Thanks David.

PR #59, opened 2015-08-26
Update index checkbounds API

mbauman:
Now formally supported and documented due to https://github.com/JuliaLang/julia/pull/11895. Note: this does not compat-ably version the change, but simply points the way to the new API.

### DIFF
--- a/src/indexing.jl
+++ b/src/indexing.jl
@@ -60,20 +60,21 @@ end
 
 # ----- Base._checkbounds ----------------------------------------------------#
 
-function Base._checkbounds{T<:Real}(sz::Int, x::Nullable{T})
-    isnull(x) ? throw(NullException()) : Base._checkbounds(sz, get(x))
+function Base.checkbounds{T<:Real}(::Type{Bool}, sz::Int, x::Nullable{T})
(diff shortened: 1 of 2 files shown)


src/indexing.jl
```diff
@@ -60,20 +60,21 @@ end
...
-    isnull(x) ? throw(NullException()) : Base._checkbounds(sz, get(x))
+function Base.checkbounds{T<:Real}(::Type{Bool}, sz::Int, x::Nullable{T})
+    isnull(x) ? throw(NullException()) : checkbounds(Bool, sz, get(x))
 end
 
-function Base._checkbounds(sz::Int, I::NullableVector{Bool})
-    length(I) == sz || throw(BoundsError())
+function Base.checkbounds(::Type{Bool}, sz::Int, I::NullableVector{Bool})
+    anynull(I) && throw(NullException())
+    length(I) == sz
```
mbauman, line 69: This changes the semantics here (beyond just the rename)… but I'm pretty sure that this is actually what you want.

johnmyleswhite, line 69: I agree that this is what's wanted.

johnmyleswhite:
Thanks, Matt!

PR #60, opened 2015-08-31
Show patch

### LINKED ISSUE #57: Custom show method doesn't handle higher-order arrays
```
julia> NullableArray(fill(1.1, (1, 1, 1)))
1x1x1 NullableArray{Float64,3}:
[:, :, 1] =
 Nullable(1.1)
```

davidagold:
I've managed to pare down the code reuse necessary to get the custom show regime working for higher dimensional `NullableArray`s. There's still a frustrating duplication of the `Base.alignment` and `Base.print_matrix_row` methods -- one to show `NullableArray` arguments of dimension <2 and one to show `SubArray` arguments with `NullableArray` parents. In fact the method bodies are precisely the same, but I haven't been able to find a method signature that properly covers both argument signatures. The difficulty is that one needs to introduce type parameters in order to cover the `SubArray` case, thereby adding an extra layer of specificity to the methods. The additional specificity causes the `Base.display` method chain to fall back on the default `Base.alignment` and `Base.print_matrix_row` methods for `NullableArray` arguments.

Regardless, this is how it looks:

``` julia
julia> using NullableArrays

julia> A = reshape(1:16, 2, 2, 2, 2);

julia> X = NullableArray(A, rand(Bool, 2, 2, 2, 2))
2x2x2x2 NullableArray{Int64,4}:
[:, :, 1, 1] =
     1      3
 #NULL  #NULL

[:, :, 2, 1] =
 5  #NULL
 6  #NULL

[:, :, 1, 2] =
 #NULL  11
    10  12

[:, :, 2, 2] =
 #NULL  15
 #NULL  16

julia> X[:, :, 1, 1]
2x2 NullableArray{Int64,2}:
     1      3
 #NULL  #NULL

julia> X[:, 1, 1, 1]
2-element NullableArray{Int64,1}:
     1
 #NULL

julia> slice(X, :, :, :, 1)
2x2x2 SubArray{Nullable{Int64},3,NullableArray{Int64,4},Tuple{Colon,Colon,Colon,Int64},4}:
[:, :, 1] =
     1      3
 #NULL  #NULL

[:, :, 2] =
 5  #NULL
 6  #NULL

julia> slice(X, :, :, 1, 1)
2x2 SubArray{Nullable{Int64},2,NullableArray{Int64,4},Tuple{Colon,Colon,Int64,Int64},4}:
     1      3
 #NULL  #NULL
```

Fixes #57

### DIFF
--- a/src/primitives.jl
+++ b/src/primitives.jl
@@ -126,16 +126,6 @@ function anynull(xs::NTuple) # -> Bool
     return anynull(collect(xs))
 end
 
-@generated function anynull{T, N, U<:NullableArray}(S::SubArray{T, N, U})
-    return quote
-        isnull = slice(S.parent.isnull, S.indexes...)
-        @nloops $N i S begin
-            (@nref $N isnull i) && (return true)
-        end
-        return false
-    end
-end
-
 # ----- allnull --------------------------------------------------------------#
 
 allnull(X::NullableArray) = all(X.isnull) # -> Bool
@@ -260,3 +250,8 @@ end
 
 # Use ready-made method for AbstractArrays or implement method specific to
 # NullableArrays, possibly for performance purposes?
+
+###
+
+Base.isnull(X::NullableArray, I::Int...) = X.isnull[I...]
+Base.values(X::NullableArray, I::Int...) = X.values[I...]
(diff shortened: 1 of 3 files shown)


johnmyleswhite:
Nice work!

PR #64, opened 2015-09-14
Add in-line docstring documentation

davidagold:
This commit also removes design notes and code section delineators in preparation for a release version.

Folks should feel free to read through things and let me know if they'd like to see changes. Or they should feel free to make whatever changes they feel are appropriate themselves.


davidagold:
For whatever reason, searches for Base methods extended in this package do not turn up in REPL help-mode searches in the current v0.4-rc version. They show up just fine in my 3-day old master which is tagged as 0.5-dev.

davidagold:
Filed an issue in the Julia repo: https://github.com/JuliaLang/julia/issues/13152

Works on master -- I'm going to merge since the problem doesn't seem to be with the present package.

PR #69, opened 2015-09-15
Include tests for `src/subarray.jl`

davidagold:


davidagold:
The "push" Travis build is failing because this PR is a few commits behind `master`. In particular, it's behind the commit that include the `src/subarray.jl` code in the main module -- hence the no method error. Since `master` has the required commit, this shouldn't be any trouble to merge -- hence the "PR" build passing. Going ahead to merge.

PR #80, opened 2015-10-01
Fix missing Union{...} deprecation

quinnj:


johnmyleswhite:
Thanks!

johnmyleswhite:
Hmm. This apparently failed, although I can't imagine why. Perhaps Travis cut it short when I merged it?

simonster:
Looks like it's because the Julia nightlies were built without libssl support (https://github.com/JuliaLang/julia/issues/13389)

simonster:
No actually it's a different error. Maybe something about the SSL config on Travis is interacting badly with libgit2, such that libgit2 doesn't have the proper root certificates?

johnmyleswhite:
Sigh. False positive testing infra...

PR #84, opened 2015-10-12
Add ==(::NullableArray, ::NullableArray) and ==(::NullableArray, ::AbstractArray)

### LINKED ISSUE #82: Definition of `==` for NullableArray doesn't seem to work
My guess is we're relying on some `AbstractArray` behavior, but the fact that we return `Nullable{Bool}` instead of `Bool` is throwing something off. To be clear, I'm not advocating for _not_ returning `Nullable{Bool}`, I think we may just need to define our own `==` or something where the semantics of `Nullable{Bool}` vs. `Bool` are currently breaking things.

``` julia
julia> a = NullableArrays.NullableArray([1])
1-element NullableArrays.NullableArray{Int64,1}:
 1

julia> b = NullableArrays.NullableArray([1])
1-element NullableArrays.NullableArray{Int64,1}:
 1

julia> a == b
ERROR: TypeError: non-boolean (Nullable{Bool}) used in boolean context
 in == at abstractarray.jl:1041

julia>
```

nalimilan:
Fixes #82.

### DIFF
--- a/src/primitives.jl
+++ b/src/primitives.jl
@@ -135,6 +135,59 @@ Returns the last entry of `X`.
 Base.endof(X::NullableArray) = endof(X.values) # -> Int
 
 @doc """
+`==(A::NullableArray, B::NullableArray)`
+`==(A::NullableArray, B::AbstractArray)`
+`==(A::AbstractArray, B::NullableArray)`
+
+Returns `Nullable(true)` if all array elements of the same rank
+are equal and none of the arrays contain missing values, `Nullable(false)`
+if two (non-missing) elements of the same rank differ, and `Nullable{Bool}()`
+otherwise.
+""" ->
+function Base.(:(==))(A::NullableArray, B::NullableArray)
+    if size(A) != size(B)
+        return Nullable(false)
+    end
+    anynull = false
+    for i in eachindex(A,B)
+        if A.isnull[i] || B.isnull[i]
+            anynull = true
+        elseif A.values[i] != B.values[i]
+            return Nullable(false)
+        end
+    end
+    if anynull
+        return Nullable{Bool}()
+    else
+        return Nullable(true)
+    end
+end
+
+function Base.(:(==))(A::NullableArray, B::AbstractArray)
+    if size(A) != size(B)
+        return Nullable(false)
+    end
+    if isa(B, Range)
+        return Nullable(false)
+    end
+    anynull = false
+    for i in eachindex(A,B)
+        if A.isnull[i]
+            anynull = true
+        elseif A.values[i] != B[i]
+            return Nullable(false)
+        end
+    end
+    if anynull
+        return Nullable{Bool}()
+    else
+        return Nullable(true)
+    end
+end
+
+Base.(:(==))(A::AbstractArray, B::NullableArray) = B == A
+
+@doc """
 
 """ ->
 function Base.find(X::NullableArray{Bool}) # -> Array{Int}
--- a/test/primitives.jl
+++ b/test/primitives.jl
@@ -149,6 +149,27 @@ module TestPrimitives
     @test endof(NullableArray(collect(1:10))) == 10
     @test endof(NullableArray([1, 2, nothing, 4, nothing])) == 5
 
+# ----- test Base.== ------------------------------------------------------#
+    x = collect(1:3)
+    xn = NullableArray(x)
+    y = [1.0, 2.0, 3.0]
+    yn = NullableArray(y, [false, false, true])
+    z = [1.1, 2.0, 3.0]
+    zn = NullableArray(z, [false, true, false])
+    @test get(x == NullableArray(x))
+    @test get(y == NullableArray(y))
+    @test get(x != NullableArray(z))
+    @test get(x != yn[1:2])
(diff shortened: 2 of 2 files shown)


nalimilan:
Let's merge this? I need it for https://github.com/JuliaStats/DataFrames.jl/pull/1008.

davidagold:
We should first define `hash` too, right?

davidanthoff:
This is probably also effected by the discussion in #85? I.e. the semantics for `==` should be the same for scalar and arrays, right?

davidagold:
>  I.e. the semantics for == should be the same for scalar and arrays, right?

I don't think so, since `NullableArray` is not itself a `Nullable` type. So `==(X::NullableArray, Y::NullableArray)` should return a `Bool`, whereas I think legitimate cases have been made for returning either a `Bool` or a `Nullable{Bool}` from `==(x::Nullable, y::Nullable)`.

davidanthoff:
> I don't think so, since `NullableArray` is not itself a `Nullable` type. So `==(X::NullableArray, Y::NullableArray)` should return a Bool, whereas I think legitimate cases have been made for returning either a `Bool` or a `Nullable{Bool}` from `==(x::Nullable, y::Nullable)`.

@davidagold I agree with you (I'm always in favor of returning a `Bool` from any comparison operator), but that is not how this PR implements it. `==(X::NullableArray, Y::NullableArray)` returns a `Nullable{Bool}` if this PR gets merged.

davidagold:
@davidanthoff Hmm, so it does. Good point. I don't think that's appropriate.

This may be unpopular, but  I think the behavior described in this PR is more appropriately expressed by `isequal`, module returning a `Nullable{Bool}`. There is something of an analogy between `NaN` and `Nullable`, and for the former, `isequal` provides a more "relaxed" behavior than `==`. So I think it is appropriate that this sort of relaxed behavior should belong to `isequal` in the case of comparing `Nullable`s as well. Indeed, Base already implements this sort of relaxed comparison behavior for `isequal(x::Nullable, y::Nullable)`, hence `isequal(X::NullableArray, Y::NullableArray)` inherits precisely the desired behavior from the `AbstractArray` interface. However, this is at the expense of having 

```
julia> isequal(Nullable(), Nullable())
true
```

@nalimilan How do you feel about using `isequal` in place of `==` in JuliaStats/DataFrames.jl#1008?

nalimilan:
The current behavior of `isequal` is indeed correct. What I'm proposing here is a definition of `==` which is consistent with that used for `Nullable` and for `AbstractArray`. Indeed, `==(::AbstractArray, ::AbstractArray)` calls `==` on each pair of elements of the arrays; that definition must be extended to return a `Nullable{Bool}` to handle missing values. Else, we would be treating nulls as `false`, which we don't currently do anywhere else.

If we were to return a `Bool` from `==(::Nullable, ::Nullable)`, then we would also have to change the present PR. But if we keep the current definition, I think the most consistent choice is to return a `Nullable{Bool}` for both.

> @nalimilan How do you feel about using isequal in place of == in JuliaStats/DataFrames.jl#1008?

I'm actually using `isequal` now, which is mostly what is needed when writing tests. But `==` for `DataFrame` recursively calls `==` on its elements, which currently triggers an error for `NullableArrays`. (That said, that function probably isn't very useful outside testing, where `isequal` is more appropriate. It's mostly needed for completeness.)

davidagold:
I see. And the behavior implemented herein is different from that of `isequal`, anyway. But won't returning `Nullable{Bool}` break the `DataFrame` comparison, then? Or have you already taken that into account? And will comparing two `DataFrame`s return a `Nullable{Bool}`?

nalimilan:
Yes, `==(::DataFrame, ::DataFrame)` would also have to return a `Nullable{Bool}`, which requires some tweaking.

davidanthoff:
> Yes, ==(::DataFrame, ::DataFrame) would also have to return a Nullable{Bool}, which requires some tweaking.

I see why having `==(::Nullable,::Nullable)` return a `Nullable{Bool}` can sometimes be useful (although I think it complicates the design way too much to be worth the hassle), but I can't come up with any scenario where comparing two `DataFrame`s or two `NullableArray`s via `==` returning `Nullable{Bool}` would actually be useful. Maybe just lack of imagination, but is there any scenario where you would compare these objects and not be in a situation where you actually want a boolean value?

nalimilan:
Honestly, I don't even know when comparing two data frames could really be useful apart from testing or checking results. My proposal just tries to preserve consistency with `==(::Nullable, ::Nullable)`, which is IMO the fundamental decision.

davidanthoff:
Yeah, but checking results might be quite a common and useful thing? I don't know, you load data and manipulate data in some way (via Query.jl or jplyr.jl), and then you change your query to be more concise and want to make sure you still produce the same result set. That is something that might be quite common, but it would require a `Bool` as a result to be useful.

nalimilan:
As I said, there's `isequal` for that already, which might be enough.

davidagold:
> then you change your query to be more concise and want to make sure you still produce the same result set. That is something that might be quite common, but it would require a Bool as a result to be useful.

That's probably not the best way to convince oneself of the equality of those queries, anyway =p

Issue #98, opened 2016-03-01
Can't construct a nullable vector from a vector of nullables


andyferris:
Case in point:

```
julia> NullableVector{Int64}([Nullable{1},Nullable{2},Nullable{3}])
ERROR: MethodError: `convert` has no method matching convert(::Type{Int64}, ::Type{Nullable{1}})
This may have arisen from a call to the constructor Int64(...),
since type constructors fall back to convert methods. 
Closest candidates are:
  call{T}(::Type{T}, ::Any) 
  convert(::Type{Int64}, ::Int8)
  convert(::Type{Int64}, ::UInt8)
  ...
 in copy! at abstractarray.jl:310
 in call at essentials.jl:56
```

This seems really unintuitive. 

I came across this when I tried to make [Tables.jl](https://github.com/FugroRoames/Tables.jl) understand about storing `Nullable` values in `NullableArrays`, where I have some convenience constructors for when, e.g., a single row (or a small number of rows) of the table are used in instantiation and my constructor needs to convert one or more `Nullable`s in a vector to a `NullableVector`.

Is there a strong reason this isn't implemented? Would you welcome a PR?

PS - Similar automatic conversion works with `Vector`:

```
julia> Vector{Int64}([1.0,2.0,3.0])
3-element Array{Int64,1}:
 1
 2
 3
```

johnmyleswhite:
I think this will work if you define the appropriate `convert` method.

andyferris:
OK, that works:

```
julia> function Base.convert{T,N}(::Type{NullableArray{T,N}},in::Array{Nullable{T},N})
           out = NullableArray{T,N}(Array(T,size(in)),Array(Bool,size(in)))
           for i = 1:length(in)
               (out.isnull[i] = in[i].isnull) ? nothing : out.values[i] = in[i].value
           end
           out
       end
convert (generic function with 549 methods)

julia> NullableArray{Int64,1}(x)
4-element NullableArrays.NullableArray{Int64,1}:
     1
     2
     3
 #NULL
```

But we'll also need the outer constructor:

```
julia> NullableArray(x)
4-element NullableArrays.NullableArray{Nullable{Int64},1}:
 Nullable(1)      
 Nullable(2)      
 Nullable(3)      
 Nullable{Int64}()
```

I will attempt a PR with both

johnmyleswhite:
You don't need an outer constructor: you just need a different `convert` method. I think you're missing two cases at least:
[...]

PR #99, opened 2016-03-02
Added ability to convert from Arrays of Nullables

andyferris:
These methods add the ability to create a `NullableArray` from data that is in an `Array{Nullable{T},N}`. Tests added.

See #98

### DIFF
--- a/src/constructors.jl
+++ b/src/constructors.jl
@@ -96,19 +96,19 @@ function NullableArray{T}(A::AbstractArray,
 end
 
 #----- Constructor #7 --------------------------------------------------------#
-
 # The following method allows for the construction of zero-element
 # NullableArrays by calling the parametrized type on zero arguments.
 # TODO: add support for dimensions arguments?
 @compat (::Type{NullableArray{T, N}}){T, N}() = NullableArray(T, ntuple(i->0, N))
 
-
 #----- Conversion from Array of Nullables #1 -----------------------------------#
 # Take an Array of Nullable objects and convert it into a NullableArray
-function Base.convert{S,T,N}(::Type{NullableArray{S,N}},in::Array{Nullable{T},N})
-   out = NullableArray{S,N}(Array(S,size(in)),Array(Bool,size(in)))
+function Base.convert{S, T, N}(::Type{NullableArray{S, N}}, in::AbstractArray{Nullable{T}, N})
+   out = NullableArray{S, N}(Array(S, size(in)), Array(Bool, size(in)))
    for i = 1:length(in)
-       (out.isnull[i] = in[i].isnull) ? nothing : out.values[i] = in[i].value
+       if !(out.isnull[i] = isnull(in[i]))
+           out.values[i] = in[i].value
+       end
    end
    out
 end
@@ -116,10 +116,10 @@ end
 #----- Conversion from Array of Nullables #2 -----------------------------------#
 # Take an Array of Nullable objects and convert it into a NullableArray
 # A slightly more convenient syntax to match the type of the input
-Base.convert{S,T,N}(::Type{NullableArray{S}},in::Array{Nullable{T},N}) = Base.convert(NullableArray{S,N},in)
+Base.convert{S, T, N}(::Type{NullableArray{S}}, in::AbstractArray{Nullable{T}, N}) = Base.convert(NullableArray{S, N}, in)
 
 #----- Constructor #8 -----------------------------------#
 # Take an Array of Nullable objects and convert it into a NullableArray
 # The constructor needs defining as it is a more specialized version of the
 # above, and thus will not fall back to convert.
-NullableArray{T,N}(in::Array{Nullable{T},N}) = Base.convert(NullableArray{T,N},in)
+NullableArray{T, N}(in::Array{Nullable{T}, N}) = Base.convert(NullableArray{T, N}, in)
--- a/src/constructors.jl
+++ b/src/constructors.jl
@@ -102,3 +102,24 @@ end
 # TODO: add support for dimensions arguments?
 @compat (::Type{NullableArray{T, N}}){T, N}() = NullableArray(T, ntuple(i->0, N))
 
+
+#----- Conversion from Array of Nullables #1 -----------------------------------#
+# Take an Array of Nullable objects and convert it into a NullableArray
+function Base.convert{S,T,N}(::Type{NullableArray{S,N}},in::Array{Nullable{T},N})
+   out = NullableArray{S,N}(Array(S,size(in)),Array(Bool,size(in)))
+   for i = 1:length(in)
+       (out.isnull[i] = in[i].isnull) ? nothing : out.values[i] = in[i].value
+   end
+   out
+end
+
+#----- Conversion from Array of Nullables #2 -----------------------------------#
+# Take an Array of Nullable objects and convert it into a NullableArray
+# A slightly more convenient syntax to match the type of the input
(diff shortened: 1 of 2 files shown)


andyferris:
Hmm... I only checked Julia 0.4. Did I cause the nightly build test to break, or was it broken already?

src/constructors.jl
```diff
@@ -103,3 +103,24 @@ end
 function Base.call{T, N}(::Type{NullableArray{T, N}})
     NullableArray(T, ntuple(i->0, N))
 end
+
+#----- Conversion from Array of Nullables #1 -----------------------------------#
+# Take an Array of Nullable objects and convert it into a NullableArray
+function Base.convert{S,T,N}(::Type{NullableArray{S,N}},in::Array{Nullable{T},N})
```
nalimilan: Please add spaces after commas as in the rest of the file.

src/constructors.jl
```diff
@@ -103,3 +103,24 @@ end
...
+   out = NullableArray{S,N}(Array(S,size(in)),Array(Bool,size(in)))
+   for i = 1:length(in)
+       (out.isnull[i] = in[i].isnull) ? nothing : out.values[i] = in[i].value
```
nalimilan: Using an assignment and then an `if` block would be more natural. `nothing` is weird here, and the expression is complex. Also use `isnull(in[i])` rather than `in[i].isnull` and `get(in[i])` rather than `in[i].value`. Storing `in[i]` in an object might remove some overhead, not sure.

src/constructors.jl
```diff
@@ -103,3 +103,24 @@ end
...
+   end
+   out
+end
+
+#----- Conversion from Array of Nullables #2 -----------------------------------#
+# Take an Array of Nullable objects and convert it into a NullableArray
+# A slightly more convenient syntax to match the type of the input
+Base.convert{S,T,N}(::Type{NullableArray{S}},in::Array{Nullable{T},N}) = Base.convert(NullableArray{S,N},in)
```
nalimilan: Maybe even add another method without any type parameter for `NullableArray`?

src/constructors.jl
nalimilan: Ah, and you don't need the `T` here and below.

test/constructors.jl
```diff
@@ -65,5 +65,16 @@ module TestConstructors
     @test isequal(NullableArray{Bool, 2}(),
                   NullableArray{Bool, 2}(Array(Bool, 0, 0)))
 
-
+    # test conversion from arrays of nullables
+    array_of_nulls = Nullable{Int}[Nullable(1), Nullable(2), Nullable(3), Nullable()]
+    @test isa(NullableArray(array_of_nulls), NullableArray{Int,1})
+    @test isequal(NullableArray(array_of_nulls), NullableArray{Int,1}([1,2,3,4],[false,false,false,true]))
```
nalimilan, line 71: Use `===` rather than isequal: that way, you'll check the actual type, not the `isequal` method. That will allow you to remove the lines using `isa`.

nalimilan:
Thanks. The Travis failure doesn't look related, but I guess it should be fixed anyway.

src/constructors.jl
andyferris: Thanks for the code review, @nalimilan. How will it work without `T`? For example, we don't have `Array{Nullable{Int,N}} <: Array{Nullable}`, and I didn't want this function to capture `Array`s that were filled with something non-`Nullable`.

src/constructors.jl
andyferris: Yes, it was weird but I it was the fastest code I could think of. I still don't see an `unsafe_get` or similar method for getting when you know it's not null, even if `in[i]` was copied locally...

test/constructors.jl
andyferris, line 71: Good tip, but `is`/`===`/`≡` for mutables checks if the memory addresses are the same, which isn't true here (my constructors copy the data so they won't end up with the same references).

For example:

```
julia> NullableArray([1,2,3]) === NullableArray([1,2,3])
false
```

src/constructors.jl
andyferris: That's in constructor #8... some earlier constructors take precedence over this one if I just define `convert`.

src/constructors.jl
nalimilan: I meant `in::Array{Nullable,N}`, but now I see that you need to keep `T` in `{S,T,N}` anyway, so you can as well keep it.

src/constructors.jl
```diff
@@ -96,10 +96,44 @@ function NullableArray{T}(A::AbstractArray,
...
+   out = NullableArray{S, N}(Array(S, size(in)), Array(Bool, size(in)))
+   for i = 1:length(in)
+       if !(out.isnull[i] = isnull(in[i]))
+           out.values[i] = in[i].value
+       end
+   end
+   out
+end
+
+#----- Conversion from Array of Nullables #1 -----------------------------------#
```
nalimilan, line 104: Duplicate of the method above?

src/constructors.jl
nalimilan: You mean that `get` gives slower code when benchmarking it? I would have thought it would be inlined and that the redundant checks for nullity would go away, but maybe that's too optimistic.

src/constructors.jl
nalimilan: OK, fine.

test/constructors.jl
nalimilan, line 71: Right. Yet you can use `==` instead of `isequal` since there are no `NaN`s.

davidagold:
I would like to follow best practices and not merge in the presence of even seemingly unrelated failures. We need to do a pretty thorough review of how Base Julia is impacting the present package. If people want to help me sleuth through this, I'd welcome it. I have finals next week and won't be able to sit down and seriously examine these issues till after then. 

EDIT: I will mention I appreciate this PR and that I don't like holding up merging it on account of what are probably unrelated failures.

davidagold:
Actually, @nalimilan @johnmyleswhite do you have any opinions about merging this once it's ready despite the state of this package on v0.5? I can put something in the README to emphasize that we don't claim to support 0.5 yet.

nalimilan:
As long as the tests pass on 0.4, it should be fine (unless the tests pass on 0.5 without this PR of course). Though there are still a few unadressed points.

davidagold:
Yes, there are. But once those are addressed, I think I'll be comfortable merging. The breakage points are not new to this PR. Thanks for weighing in, @nalimilan .

johnmyleswhite:
Agree with Milan that you should focus on making sure tests pass on 0.4. Targeting 0.5 gives you a moving target.

davidagold:
Indeed. Okay, PRs need pass tests only on 0.4. Thank you for your thoughts, John.

nalimilan:
@andyferris Would you mind rebasing your PR? We were quite close from merging when the discussion stopped.

andyferris:
Sorry, yes I have been distracted by many things (moving house, etc) and this one went on the back-burner. I'll try have a look tomorrow!

andyferris:
OK, rebased and removed an duplicated function.  Tests pending, but could be good to go?

nalimilan:
Thanks! Looks good to me, if @davidagold wants to merge. Though you should probably squash the last commit.

davidagold:
Yes, would you please squash those into one commit? I'd greatly appreciate that =) Otherwise, this looks good to go.

nalimilan:
Thanks, @andyferris! I've squashed them directly using the new GitHub support for that.

Issue #100, opened 2016-04-02
NullableArrays broken in v0.5 by 15071


ScottPJones:
https://github.com/JuliaLang/julia/pull/15071 removed the `index_generate` function, which is used by `NullableArrays`.
I've been trying to get `NullableArrays` working again in v0.5, and this looks like the last stumbing block.

ScottPJones:
cc: @mbauman Do you have any suggestions as to how to proceed here? Thanks!

mbauman:
SubArrays are much easier to work with now.  No more need for generated functions.  You can recompute the indices into the parent array with `Base.reindex`.  I think the functions in subarray.jl can be simplified down to just `getindex(V.parent.isnull, Base.reindex(V, V.indexes, I)...)`. That should give you enough to get you on the right path.

nalimilan:
@ScottPJones Can you make a PR with these changes?

ScottPJones:
@nalimilan I hadn't realized the issue wasn't automatically closed, but David Gold already merged my fixes in 8 days ago.

nalimilan:
OK, thanks.

PR #101, opened 2016-04-03
Fix issue #100, make work again with v0.5

ScottPJones:
This fixes deprecations due to removal of `call` overloading,
and also fixes a problem with handling subarrays,
due to https://github.com/JuliaLang/julia/pull/15071 getting rid of the unexported `index_generate` function.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,2 +1,3 @@
 julia 0.4-
+Compat
 Reexport
--- a/src/NullableArrays.jl
+++ b/src/NullableArrays.jl
@@ -1,34 +1,35 @@
 module NullableArrays
 
-    using Compat
-    using Reexport
-    @reexport using Base.Cartesian
+using Compat
+using Reexport
+@reexport using Base.Cartesian
 
-    export NullableArray,
-           NullableVector,
-           NullableMatrix,
+export NullableArray,
+       NullableVector,
+       NullableMatrix,
 
-           # Macros
+       # Macros
 
-           # Methods
-           dropnull,
-           anynull,
-           allnull,
-           head,
-           nullify!,
-           padnull!,
-           padnull,
-           tail
+       # Methods
+       dropnull,
+       anynull,
+       allnull,
+       head,
+       nullify!,
+       padnull!,
+       padnull,
+       tail
 
-    include("typedefs.jl")
-    include("constructors.jl")
-    include("primitives.jl")
-    include("indexing.jl")
-    include("map.jl")
-    include("nullablevector.jl")
-    include("operators.jl")
-    include("broadcast.jl")
-    include("reduce.jl")
-    include("show.jl")
-    include("subarray.jl")
+include("typedefs.jl")
+include("constructors.jl")
+include("primitives.jl")
+include("indexing.jl")
+include("map.jl")
+include("nullablevector.jl")
+include("operators.jl")
+include("broadcast.jl")
+include("reduce.jl")
+include("show.jl")
+include("subarray.jl")
+
 end
--- a/src/NullableArrays.jl
+++ b/src/NullableArrays.jl
@@ -1,5 +1,6 @@
 module NullableArrays
 
+    using Compat
     using Reexport
     @reexport using Base.Cartesian
 
--- a/src/constructors.jl
+++ b/src/constructors.jl
@@ -100,6 +100,5 @@ end
 # The following method allows for the construction of zero-element
 # NullableArrays by calling the parametrized type on zero arguments.
 # TODO: add support for dimensions arguments?
-function Base.call{T, N}(::Type{NullableArray{T, N}})
-    NullableArray(T, ntuple(i->0, N))
-end
+@compat (::Type{NullableArray{T, N}}){T, N}() = NullableArray(T, ntuple(i->0, N))
+
--- a/src/operators.jl
+++ b/src/operators.jl
@@ -104,18 +104,18 @@ end
 
 ## Lifted functors
 
-function Base.call{S1, S2}(::Base.MinFun, x::Nullable{S1}, y::Nullable{S2})
+@compat function (::Base.MinFun){S1, S2}(x::Nullable{S1}, y::Nullable{S2})
     if isbits(S1) & isbits(S2)
         return Nullable(Base.scalarmin(x.value, y.value), x.isnull | y.isnull)
     else
         error()
     end
 end
-
-function Base.call{S1, S2}(::Base.MaxFun, x::Nullable{S1}, y::Nullable{S2})
+@compat function (::Base.MaxFun){S1, S2}(x::Nullable{S1}, y::Nullable{S2})
     if isbits(S1) & isbits(S2)
         return Nullable(Base.scalarmax(x.value, y.value), x.isnull | y.isnull)
     else
         error()
     end
 end
+
--- a/src/subarray.jl
+++ b/src/subarray.jl
@@ -1,3 +1,6 @@
+if VERSION >= v"0.5.0-dev+2718"
+    include("subarray0_5.jl")
+else
 const unsafe_getindex = Base.unsafe_getindex
 
 @generated function Base.isnull{T,N,P<:NullableArray,IV,LD}(V::SubArray{T,N,P,IV,LD}, I::Int...)
@@ -32,6 +35,7 @@ end
         $exhead
         Base.getindex(V.parent.values, $(idxs...))
     end
+end
 end
 
 @generated function anynull{T, N, U<:NullableArray}(S::SubArray{T, N, U})
--- a/src/subarray0_5.jl
+++ b/src/subarray0_5.jl
@@ -1,0 +1,36 @@
+typealias NullableSubArray{T,N,P<:NullableArray,IV,LD} SubArray{T,N,P,IV,LD}
+
+@inline function Base.isnull(V::NullableSubArray, I::Int...)
+    @boundscheck checkbounds(V, I...)
+    @inbounds return V.parent.isnull[Base.reindex(V, V.indexes, I)...]
+end
+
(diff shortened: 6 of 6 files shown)


ScottPJones:
@andreasnoack @ElOceanografo Could somebody from JuliaStats review this and possibly merge this?
@mbauman Does my change based on your suggestion look correct for subarrays?  Thanks!!!

src/subarrayn.jl
```diff
@@ -0,0 +1,25 @@
+const unsafe_getindex = Base.unsafe_getindex
+
+function Base.isnull{T,N,P<:NullableArray,IV,LD}(V::SubArray{T,N,P,IV,LD}, I::Int...)
+    @Base._inline_meta
```
mbauman: Just use `@inline function...`. This private macro only exists for bootstrapping.

src/subarrayn.jl
```diff
@@ -0,0 +1,25 @@
...
+end
+
+function Base.values{T,N,P<:NullableArray,IV,LD}(V::SubArray{T,N,P,IV,LD}, I::Int...)
+    @Base._inline_meta
+    @boundscheck checkbounds(V, I...)
+    @inbounds r = V.parent.values[Base.reindex(V, V.indexes, Base.to_indexes(I...))...]
+    r
+end
+
+@generated function anynull{T, N, U<:NullableArray}(S::SubArray{T, N, U})
```
mbauman: You didn't make any changes to this fubction, did you? Let's not make unnecessary duplications across versions here. I'd just put the version switch in the existing SubArray file.

mbauman:
There will be a performance regression for linear indexing with linear fast SubArrays here. That can be avoided with an additional method that uses the stride computation as in https://github.com/JuliaLang/julia/blob/57c53afba29a06493a2d993fb571a992d0ad733a/base/subarray.jl#L151.

src/subarrayn.jl
```diff
@@ -0,0 +1,25 @@
+const unsafe_getindex = Base.unsafe_getindex
+
+function Base.isnull{T,N,P<:NullableArray,IV,LD}(V::SubArray{T,N,P,IV,LD}, I::Int...)
+    @Base._inline_meta
+    @boundscheck checkbounds(V, I...)
+    @inbounds r = V.parent.isnull[Base.reindex(V, V.indexes, Base.to_indexes(I...))...]
```
mbauman: Don't use `to_indexes`. They're already all Int.

nalimilan:
Could you move the unrelated indentation changes to a separate commit?

src/subarrayn.jl
```diff
@@ -0,0 +1,25 @@
+const unsafe_getindex = Base.unsafe_getindex
```
nalimilan: This doesn't appear to be used anywhere.

ScottPJones: That was just in the original code, so I left it in, didn't know if it was necessary. I'll remove it.

src/subarrayn.jl
ScottPJones: Originally I was running into problems with still getting because the parser was giving warnings, and the first way I tried to prevent that was to use separate files (it turned out that using `@compat` was enough to solve those problems).  Fixed now.

src/subarrayn.jl
ScottPJones: Turns out, to_indexes for some reason is necessary, if I don't use it, it gets an error, saying that it can't find a method of reindex that matches.  Any idea why?

src/subarrayn.jl
ScottPJones: It gives strange error messages under v0.4.5 (fine under v0.5) when I try to combine them:

``` julia
ERROR: LoadError: LoadError: wrong number of arguments
 in include at /Applications/Julia-0.4.5.app/Contents/Resources/julia/lib/julia/sys.dylib
 in include_from_node1 at /Applications/Julia-0.4.5.app/Contents/Resources/julia/lib/julia/sys.dylib
 in include at /Applications/Julia-0.4.5.app/Contents/Resources/julia/lib/julia/sys.dylib
 in include_from_node1 at /Applications/Julia-0.4.5.app/Contents/Resources/julia/lib/julia/sys.dylib
 in require at /Applications/Julia-0.4.5.app/Contents/Resources/julia/lib/julia/sys.dylib
while loading /Users/scott/.julia/v0.4/NullableArrays/src/subarray.jl, in expression starting on line 102
while loading /Users/scott/.julia/v0.4/NullableArrays/src/NullableArrays.jl, in expression starting on line 33
```

The error corresponds to the line after the `end` from the `if VERSION < v"0.5.0-dev+2718"` ... `else` ... `end`.
I have no idea what is causing this issue in v0.4.5, a bug in the compiler maybe?

src/subarrayn.jl
mbauman: Are you splatting `I`?  Don't splat `I`.

src/subarrayn.jl
mbauman: `@boundscheck` means something different in 0.4, where it takes two arguments.  That's probably the issue.  I'm not sure if Compat has this handled yet, but you should (eventually) be able to to `@compat @boundscheck …`.

ScottPJones: OK, for the moment, I put the new stuff into `subarraynew.jl`, and included it in `subarray.jl`.  That's the only way I could think of to completely avoid parsing (and executing the macros) the code for v0.5.  Is there a better technique that is recommended?  Thanks again for the review!

src/subarrayn.jl
ScottPJones: Turns out it is necessary for the v0.4 code.

ScottPJones:
@mbauman @nalimilan I think this is all set to go now.  Thanks to both of you for the review (learned some neat stuff about the new subarray changes to boot!)

src/subarraynew.jl
```diff
@@ -0,0 +1,36 @@
+typealias NullableSubArray{T,N,P<:NullableArray,IV,LD} SubArray{T,N,P,IV,LD}
+
+@inline function Base.isnull(V::NullableSubArray, I::Real...)
+    @boundscheck checkbounds(V, I...)
+    @inbounds return V.parent.isnull[Base.reindex(V, V.indexes, Base.to_indexes(I...))...]
```
mbauman: Keep the signature the same between 0.4 and 0.5, and don't use `to_indexes`.  Widening the signature is an unrelated change from what's required here, and it introduces a feature difference between code that's meant to be functionally the same.

src/subarrayn.jl
mbauman: Ah, yes, I see that hasn't been done yet (https://github.com/JuliaLang/Compat.jl/issues/182).  This seems to be an okay solution for now.

mbauman:
Besides https://github.com/JuliaStats/NullableArrays.jl/pull/101#discussion_r58319542 the SubArray commit looks good to me.  I didn't review the other commits.

src/subarraynew.jl
ScottPJones: It didn't work without `to_indexes`.  I removed them, and then got an method error about `reindex`.

mbauman: What did you try, and what error did you get?  Don't splat `I`.  Just `V.parent.isnull[Base.reindex(V, V.indexes, I)...]`.

ScottPJones: Never mind - the problem was an issue with splatting, I needed to remove both the to_index call _and_ the splatting.

ScottPJones:
OK, done now, fixed all of the subarray issues, the other issue was simply using `@compat` in 3 places to handle the deprecation of `call` overloading.

src/subarraynew.jl
El-Zurdo-Sordo: ...and I didn't see your note until after I'd finally figured it out myself. duh!

mbauman:
LGTM.  Thanks @ScottPJones.

src/subarraynew.jl
nalimilan: Why do you introduce these `to_index` calls in one commit, and remove them in the next one?

El-Zurdo-Sordo: @nalimilan They were necessary to have the subscripts take `Real`, which would be more consistent with the code in `Base/subarray.jl`, however, I removed them based on @mbauman review comment above.
I do feel that it's incorrect to have these act differently than plain SubArrays, however, that could be fixed in a separate PR if people agree.

ScottPJones: @nalimilan They were necessary to have the subscripts take Real, which would be more consistent with the code in Base/subarray.jl, however, I removed them based on @mbauman review comment above.
I do feel that it's incorrect to have these act differently than plain SubArrays, however, that could be fixed in a separate PR if people agree.

kmsquire: Presumably the commits will be squashed, removing the the addiction and
subtraction of that feature?  I believe this is what @naliman was referring
to.

ScottPJones: @kmsquire I was asked to keep the fixing of the indentation as a separate commit, and I think the fixes related to adding Compat should probably be in a separate commit from the fixes related to JuliaLang/julia#15071, but I'll squash out that commit.  Does that sound reasonable?

kmsquire: Perfect

nalimilan: Yes, that was what I meant.

ScottPJones:
Bump!  Anybody can merge this fix around?  Thanks!

davidagold:
@ScottPJones Thank you very much for this =) I have just one request -- could you please change the filename`subarraynew.jl` to something more specific, like `subarray0_5.jl`? Does that seem like a reasonable request?

ScottPJones:
@davidagold Very reasonable! (I should have thought of that myself! 😳)  Thank _you_ for the nice package!

PR #105, opened 2016-04-30
Fix functor-related deprecation warnings on Julia 0.5

nalimilan:
Also move mapreduce(::Any, ::Union{typeof(&), typeof(|)}, ::NullableArray, ...)
before the more general method to avoid ambiguity warning.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,3 +1,3 @@
 julia 0.4-
-Compat
+Compat 0.7.14
 Reexport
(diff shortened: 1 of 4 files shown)


davidagold:
Is there a specific issue/PR in base that could be referenced here to give some context?

nalimilan:
See https://github.com/JuliaLang/julia/pull/15804.

davidagold:
Thank you, Milan.

PR #106, opened 2016-05-07
Remove head() and tail()

nalimilan:
These functions do not belong to NullableArrays, and they conflict
with DataArrays. They should live in Julia Base or a utilities
package.

I'm currently trying to port DataFrames to NullableArrays, and this creates conflict with DataArrays until I remove it completely.


nalimilan:
Failure on 0.4 looks unrelated.

PR #107, opened 2016-05-08
Fix convert(NullableArray, ::AbstractArray)

nalimilan:
Instead of defining a NullableArray(a::AbstractArray) constructor, define
convert() methods and rely on the constructor fallback to convert(). This ensures
both ways are supported. Move all convert(NullableArray, ...) methods to
constructors.jl.

Also refactor tests to make them more systematic. Remove numbering of constructors,
since it's annoying when adding or removing one: refer to them using signatures instead.

### DIFF
--- a/src/constructors.jl
+++ b/src/constructors.jl
@@ -1,23 +1,14 @@
 
 # ----- Outer Constructors -------------------------------------------------- #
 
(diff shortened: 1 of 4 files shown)


johnmyleswhite:
Looks good to me

PR #104, opened 2016-04-30
Update show methods to Julia 0.5

nalimilan:
Since JuliaLang/julia#15928, Nullable() is printed as #NULL, which
means we no longer need to override show methods from Base.
This fixes breakage in current Julia 0.5.
Also fix a bug in show(), where NullableArray{T,N} was printed as the
element type prefix: add parentheses to make it clear this is the type
of the container.

Fixes https://github.com/JuliaStats/NullableArrays.jl/issues/103. The diff is messy, but all I did is move the bottom of the file under the `else`, and add `show()` for 0.5. This means the day we'll drop support for 0.4, the file will be almost empty. Though since the printing code is being refactored in Base, we may well have to adjust this again; at some point, when things will have stabilized, we should try to rely only on public APIs, or copy `show_delim_array` (which is the only private function we use now).

There's a small regression/difference as regards the alignment of `#NULL`: it was previously aligned on the right, and it's now on the left. If that's considered a bad thing, we should make a simple PR against Base to add `alignment(::Nullable)`.

PR #108, opened 2016-05-24
Update show() to adapt to latest changes in Julia 0.5

nalimilan:
With latest git master, no method needs to be redefined at all to get
correct printing. This drops support for older 0.5 prereleases (i.e.
printing won't be perfect there), as changes have been too substantial
to be tracked.

Cf. previous changes in https://github.com/JuliaStats/NullableArrays.jl/pull/104.


davidagold:
Travis failure on nightly is unrelated. I think it's about time this got merged.

PR #110, opened 2016-05-26
Fix unshift! docstring and faster varargs

bkamins:
Old version of `Base.unshift!(X::NullableVector, vs...)` was very slow for large number of varargs (especially on the first run for a given set of types of arguments) as Julia has to generate many intermediate function signatures. Here is an example:

```
x = NullableArray(Int64,0)
@time unshift!(x,(1:10000)...);
```

Now it works much faster than the old version.

Additionally small fix in unshift! method signature.

### DIFF
--- a/src/nullablevector.jl
+++ b/src/nullablevector.jl
@@ -66,18 +66,6 @@ function Base.unshift!(X::NullableVector, v) # -> NullableVector{T}
 end
 
 @doc """
-`unshift!(X::NullableVector, vs...)`
-
-Insert multiple values `vs` at the beginning of `X` and return `X`.
-""" ->
-function Base.unshift!(X::NullableVector, vs...)
-    for i in endof(vs):-1:1
-        unshift!(X, vs[i])
-    end
-    return X
-end
-
-@doc """
 `shift!{T}(X::NullableVector{T})`
 
 Remove the first entry from `X` and return it as a `Nullable` object.
@@ -197,8 +185,28 @@ collection to `append!`.
 function Base.append!(X::NullableVector, items::AbstractVector)
     old_length = length(X)
     nitems = length(items)
+    items_copy = convert(typeof(X), items)
     resize!(X, old_length + nitems)
-    X[old_length + 1:end] = items[1:nitems]
+    X[old_length+1:end] = items_copy
+    return X
+end
+
+@doc """
+`prepend!(X::NullableVector, items::AbstractVector)`
+
+Add the elements of `items` to the beginning of `X`.
+
+Note that `prepend!(X, [1, 2, 3])` is equivalent to `unshift!(X, 1, 2, 3)`,
+where the items to be added to `X` are passed individually to `unshift!` and as a
+collection to `prepend!`.
+""" ->
+function Base.prepend!(X::NullableVector, items::AbstractVector)
+    old_length = length(X)
+    nitems = length(items)
+    items_copy = convert(typeof(X), items)
+    ccall(:jl_array_grow_beg, Void, (Any, UInt), X.values, nitems)
+    ccall(:jl_array_grow_beg, Void, (Any, UInt), X.isnull, nitems)
+    X[1:nitems] = items_copy
     return X
 end
 
@@ -220,7 +228,7 @@ Insert `front` null entries at the beginning of `X` and add `back` null entries
 at the end of `X`. Returns `X`.
 """ ->
 function padnull!{T}(X::NullableVector{T}, front::Integer, back::Integer)
-    unshift!(X, fill(Nullable{T}(), front)...)
+    prepend!(X, fill(Nullable{T}(), front))
     append!(X, fill(Nullable{T}(), back))
     return X
 end
--- a/src/nullablevector.jl
+++ b/src/nullablevector.jl
@@ -55,7 +55,7 @@ function Base.unshift!(X::NullableVector, v::Nullable) # -> NullableVector{T}
 end
 
 @doc """
-`unshift!(X::NullableVector, v::Nullable)`
+`unshift!(X::NullableVector, v)`
 
 Insert a value `v` at the beginning of `X` and return `X`.
 """ ->
@@ -71,7 +71,10 @@ end
 Insert multiple values `vs` at the beginning of `X` and return `X`.
 """ ->
 function Base.unshift!(X::NullableVector, vs...)
-    return unshift!(unshift!(X, last(vs)), vs[1:endof(vs)-1]...)
+    for i in endof(vs):-1:1
+        unshift!(X, vs[i])
+    end
+    return X
 end
 
 @doc """
--- a/src/nullablevector.jl
+++ b/src/nullablevector.jl
@@ -185,9 +185,8 @@ collection to `append!`.
 function Base.append!(X::NullableVector, items::AbstractVector)
     old_length = length(X)
     nitems = length(items)
-    items_copy = convert(typeof(X), items)
     resize!(X, old_length + nitems)
-    X[old_length+1:end] = items_copy
+    copy!(X, length(X)-nitems+1, items, 1, nitems)
     return X
 end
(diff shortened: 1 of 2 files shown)


nalimilan:
Actually, I think we should only define the one-argument method, and rely on the fallback definition provided by Base for the general case. Same for `push!`. Would you do that?

(BTW, it's generally not a good idea to splat collections with many elements. In the present case, better use `prepend!`.)

bkamins:
I will work on it. Seems (see JuliaLang/julia#16631) that it is worth to consider adding `append!` and `prepend!` definitions for `NullableArrays`.

bkamins:
Updated PR has the following changes:
1. removed `unshift!` definition for more than one element (it falls back to default definition)
2. fixed `append!` as appending array of wrong type raised error but corrupted original NullableArray (array was grown first and assigned to next)
3. defined `prepend!'
4. changed `padnull!` to use `prepend!`

The change to `apppend!` calls `convert` to make sure that the supplied array can be appended. This currently fails in some cases, but the problem is with definition of `convert` (but I understand that this should be fixed soon). In particular `convert` fails to convert `[1, Nullable()]` to `NullableArray{Int64}` and it should.

The definition of `prepend!` calls Julia internal C function so I do not know if this is acceptable way to do this.

nalimilan:
> The change to apppend! calls convert to make sure that the supplied array can be appended.

That's a good idea in theory, but Julia Base doesn't do this for `Array`, so I think we shouldn't do that for now either. See https://github.com/JuliaLang/julia/issues/15868.

> In particular convert fails to convert [1, Nullable()] to NullableArray{Int64} and it should.

This works in 0.5 now: you get an `Array{Nullable{Int}}`, which can be converted to a `NullableArray` if needed.

src/nullablevector.jl
```diff
@@ -194,8 +185,28 @@ collection to `append!`.
...
     old_length = length(X)
     nitems = length(items)
+    items_copy = convert(typeof(X), items)
     resize!(X, old_length + nitems)
-    X[old_length + 1:end] = items[1:nitems]
+    X[old_length+1:end] = items_copy
+    return X
+end
+
+@doc """
```
nalimilan, line 193: We should really remove all docstrings which are redundant with Julia Base (and make PRs to improve them were needed). But for now probably better follow the current convention.

test/nullablevector.jl
```diff
@@ -152,9 +155,21 @@ module TestNullableVector
...
                   NullableArray([2:14...]))
-    @test isequal(append!(X, [Nullable(15), Nullable()]),
+    @test isequal(append!(X, [Nullable(15), Nullable{Int64}()]),
                   NullableArray([2:16...], vcat(fill(false, 14), true)))
 
+    #--- test Base.prepend!
+
+    # Base.prepend!(X::NullableVector, items::AbstractVector)
+
+    XX = NullableArray([3:12...])
```
nalimilan: Just call it `X` again, this variable is used several times anyway in this file.

test/nullablevector.jl
```diff
@@ -152,9 +155,21 @@ module TestNullableVector
...
+    @test isequal(prepend!(XX, [1, 2]),
+                  NullableArray([1:12...]))
+    @test isequal(prepend!(XX, [Nullable(-1), Nullable(0)]),
+                  NullableArray([-1:12...]))
+    @test isequal(prepend!(XX, [Nullable{Int64}(), Nullable(-2)]),
+                  NullableArray([-3:12...], vcat(true, fill(false, 15))))
```
nalimilan: Even on 0.4, you can write this as `NullableArray(Nullable{Int}[Nullable(), -2:12...])`, which is more readable.

test/nullablevector.jl
```diff
@@ -152,9 +155,21 @@ module TestNullableVector
                   NullableArray([2:12...]))
     @test isequal(append!(X, [Nullable(13), Nullable(14)]),
                   NullableArray([2:14...]))
-    @test isequal(append!(X, [Nullable(15), Nullable()]),
+    @test isequal(append!(X, [Nullable(15), Nullable{Int64}()]),
```
nalimilan: `Int64` should be `Int`. But did the test fail before this?

test/nullablevector.jl
```diff
@@ -152,9 +155,21 @@ module TestNullableVector
...
                   NullableArray([2:16...], vcat(fill(false, 14), true)))
 
+    #--- test Base.prepend!
+
+    # Base.prepend!(X::NullableVector, items::AbstractVector)
+
+    XX = NullableArray([3:12...])
+    @test isequal(prepend!(XX, [1, 2]),
+                  NullableArray([1:12...]))
```
nalimilan: `NullableArray(1:12)` is enough.

src/nullablevector.jl
```diff
@@ -194,8 +185,28 @@ collection to `append!`.
...
+
+Note that `prepend!(X, [1, 2, 3])` is equivalent to `unshift!(X, 1, 2, 3)`,
+where the items to be added to `X` are passed individually to `unshift!` and as a
+collection to `prepend!`.
+""" ->
+function Base.prepend!(X::NullableVector, items::AbstractVector)
+    old_length = length(X)
+    nitems = length(items)
+    items_copy = convert(typeof(X), items)
+    ccall(:jl_array_grow_beg, Void, (Any, UInt), X.values, nitems)
```
nalimilan, line 205: Looks like we need a new function in Julia to allow doing this in pure Julia at some point.

bkamins:
@nalimilan I have taken into account all your comments (hopefully :)).

In particular I fixed other parts of tests using the style you recommended in my code.

I also tried to make `append!` and `unshift!` work exactly like they work in Julia Base.

However, I have a problem with the fact that they are not atomic. I understand the arguments given in previous threads, but when we use `NullableArrays` most of the time we have to perform conversion to NullableArray anyway in the background - only after the array is grown. So in my opinion the time saved is not really worth the unsafety it introduces (but I am not an expert).

Imagine what fraction of Julia users would correctly say what is the result of the following code after reading documentation of `copy!` (which is used inside `append!` and `unshift!`):

```
using NullableArrays
x = NullableArray(1:10)
copy!(x, 2, x, 1, 9)
```

nalimilan:
Thanks, looks good to me. I agree the behavior of `append!` on failure isn't great, but this should be discussed and fixed in Base. The situation is going to be even worse if each package chooses a different solution.

The behavior of `copy!` in case of aliasing is also annoying since it's different from standard `Array`. Could you file an issue in Julia, showing the discrepancy with the `AbstractArray` method that `NullableArray` uses?

src/nullablevector.jl
```diff
@@ -195,7 +186,29 @@ function Base.append!(X::NullableVector, items::AbstractVector)
...
+Add the elements of `items` to the beginning of `X`.
+
+Note that `prepend!(X, [1, 2, 3])` is equivalent to `unshift!(X, 1, 2, 3)`,
+where the items to be added to `X` are passed individually to `unshift!` and as a
+collection to `prepend!`.
+""" ->
+function Base.prepend!(X::NullableVector, items::AbstractVector)
+    old_length = length(X)
+    nitems = length(items)
+    ccall(:jl_array_grow_beg, Void, (Any, UInt), X.values, nitems)
```
nalimilan, line 205: This sounds like a feature which should be supported in Julia, for example via `resize!(X, n, beginning::Bool=false)`. For now it's OK to use `ccall`, but would you be willing to make a PR against Julia to add this?

davidagold:
@bkamins Thank you for this work, and my apologies it's taken this long for me to review this PR. 

The failure on nightly is, as per usual, `broadcast` related. I'm inclined to merge this. @nalimilan do you have any more comments?

nalimilan:
No further comments. Thanks!

davidagold:
Thanks Milan! Also, welcome back =)

Issue #112, opened 2016-06-10
Add a "memory reference" field?


quinnj:
Hey all,

I've been doing a round of updates on the DataStreams packages lately (big users of this package), as well as developing the new Feather.jl package for reading/writing feather files.

In my journeys, I also started playing with a new Arrow.jl package based on the Apache Arrow project which aims to define a common in-memory "dataframe" layout (or least columnar data layout). While I think an Arrow.jl project would be awesome, the more I've dug into it, I've realized it's not a trivial implementation and even shares/overlaps with a lot of the work that's happened in this package.

For now, I want to re-invest in NullableArrays as the underlying storage, but have realized a slight limitation. A lot of workflows involve getting pointers to memory regions (mmapped or otherwise) and "reinterpreting" values out of the memory region. It's obviously ideal to not need to always copy the memory region, but just hold a reference back to it. 

The limitation is the need to hold a reference back to this memory region, while the `x.isnull` and `x.values` arrays are reinterpreted from offsets into the region.

I'd like to propose adding a single `ref::Vector{UInt8}` field to `NullableArray` that could be used to potentially hold a reference to a memory region. It would be almost completely invisible to the user, but there if needed by packages like CSV/Feather where a reference is needed.

I'm happy to put in the leg work here if everyone's on-board.

nalimilan:
I don't fully understand your proposal. What's the use of that memory region if you also store `x.values`? In what occasions would you use `x.ref`? Couldn't that field be stored elsewhere, e.g. in a data frame-like structure which would group `NullableArrays` together?

quinnj:
That's actually what I'm doing now in https://github.com/JuliaDB/DataStreams.jl/blob/133cccdf6a2515f3e881d58a16f706606dd7b550/src/DataStreams.jl#L184 for Data.Table, but this isn't ideal since someone could still potentially grab the NullableArrays underneath the Data.Table, drop the reference to the parent Data.Table and now the underlying NullableArrays data can potentially be invalid. It's better to keep the reference as close to the data as possible.

So more concretely, my proposal is to have the NullableArray definition be

``` julia
immutable NullableArray{T, N} <: AbstractArray{Nullable{T}, N}
    values::Array{T, N}
    isnull::Array{Bool, N}
    ref::Vector{UInt8}
end
```

The idea is that you don't/wouldn't actually need to use `x.ref` in any case, it's just there to hold a reference so that the underlying memory region doesn't get freed/unmapped.

[...]

PR #114, opened 2016-06-11
Add a ref field to NullableArray

quinnj:
Allows for easy interop in shared memory scenarios where a reference to a memory region is needed to prevent the `x.values` array from being invalidated. Implements #112.

I'd love to try this out, and finish another round of updates based on it for the DataStreams network (CSV, SQLite, ODBC, Feather).

We can even not tell anyone if desired, then I'm the only one who has breakage if we choose to deprecate at some point :)

### DIFF
--- a/src/typedefs.jl
+++ b/src/typedefs.jl
@@ -21,15 +21,17 @@ reusing operations that only work on arrays without any null values.
 immutable NullableArray{T, N} <: AbstractArray{Nullable{T}, N}
     values::Array{T, N}
     isnull::Array{Bool, N}
+    # extra field for potentially holding a reference to a parent memory block
+    # (think mmapped file, for example) that `values` is actually derived from
+    parent::Vector{UInt8}
 
-    function NullableArray(d::AbstractArray{T, N}, m::Array{Bool, N})
+    function NullableArray(d::AbstractArray{T, N}, m::Array{Bool, N}, parent::Vector{UInt8}=Vector{UInt8}())
         if size(d) != size(m)
             msg = "values and missingness arrays must be the same size"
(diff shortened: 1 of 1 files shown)


src/typedefs.jl
```diff
@@ -21,13 +21,14 @@ reusing operations that only work on arrays without any null values.
 immutable NullableArray{T, N} <: AbstractArray{Nullable{T}, N}
     values::Array{T, N}
     isnull::Array{Bool, N}
+    ref::Vector{UInt8}
```
nalimilan: Maybe find a different name: we'll likely have to add the same hack to CategoricalArrays, which have a `refs` field. `parent`, `memblock`, `memregion`, `buffer`?

quinnj:
Updated :)

nalimilan:
Looks good to me.

quinnj:
Failing master is unrelated (more due to the mass of changes for 0-indexing [here](https://github.com/JuliaLang/julia/commit/dab5c8d8744461888bd5f1132deceae09e248af7#diff-8e941cd1878722cc12a54e01537b8e18L90)). I have some fixes for failing master tests on the way.

quinnj:
Bump. It'd be great to have this in. It'd also be nice to be added as a contributor if no one is seriously opposed. I promise to still do PRs for everything, it'd just be nice to do my own branches for test fixes and what not.

nalimilan:
Ah, I thought you had the permission to merge, as a member of JuliaStats. I'm not sure how this works, but I've just added you to the list of contributors to NullableArrays using my brand-new owner role.

quinnj:
Looks like it works 😏

PR #115, opened 2016-06-12
[WIP/RFC] functional operations on nullables as collections

TotalVerb:
This implements Scala-style `map` and `filter`. I did not implement Scala's fold semantics, because I consider that to be a mistake, since it is inconsistent with the regular `foldr` and `foldl`, which I did keep. Since `map` is not type-stable, I also implemented `broadcast` as a type-stable `map`. `broadcast` also preserves the broadcasting capability of the version for arrays.

I also implemented x[] and x[1] to replace `get(x)`. Finally, I added in the iteration interface, which allows imperative-style for loops as well as functional-style higher order functions.

John presented several options for propagation in this [old mailing list post](https://groups.google.com/forum/#!msg/julia-dev/WD7-vQeweJE/yGBFU2E2b2cJ), which I quote:

```
As I see it, progagation can work in a couple of ways:

(1) Always off: you can't propagate nullability without explicitly constructing a Nullable object.
(2) Opt-in at the call-site: you can propagate nullability by using something like map(f, Nullable).
(3) Opt-in near the call-site: you can propagate nullability within a block of code using something like @propagate begin x = Nullable(1) + Nullable(2); sin(x) end
(4) Opt-in at the definition site: you can propagate nullability by annotating a function as propagating in the same way that we annotate functions as vectorizable.
(5) Always on: you can't prevent propagation without explicitly checking for nullability before calling any function.
```

The implementation of `broadcast` here doubles as an implementation of (2). This is simply a nice result of the algebraic properties of broadcasting. With new broadcast syntactic sugar, (2) is no longer very verbose. I think alternative syntaxes like ? are hence no longer required.

---

Notes: (optional, sorry for the long PR)
- I avoided implementing the Haskell-style monadic functions, because I don't think Julia has any monads and so that would be a big departure. The Scala-style functions feel more typical in a Julia environment.
- I haven't optimized for performance yet. Obviously some functions will need to be specialized. But since 0.5 allows specialization on functions, I am sure that this syntax can be made as fast as any manual lowering.
- Over the course of this PR I discovered some further nasty things with nullable, such as how `get(Nullable(), 1)` doesn't work (it expects a `Union{}` second argument?!) and how equality comparison is a bit broken (try `Nullable() == Nullable()`). I am new to Nullables so please excuse if there is a reason for this behaviour, but I found it heavily unintuitive.
- This is only turned on on v0.5- right now. It can probably be extended to 0.4 (I don't see why not) but performance will be terrible and the shorthand `.` broadcast syntax will not be available.
- I would prefer to deprecate `get(u)` in favour of `u[]`. The former syntax seems strange to me, as all other uses of `get` in Julia include a default value. I don't know if this is a popular viewpoint; my background in nullables comes more from Lisp (where `(cons x nil)` can be used as `Nullable(x)`) and Scala than from numerical computing languages.
- Mathematically, I'm treating nullables akin to infinite-dimensional square arrays here. (More precisely, a Nullable type is an union of infinite-order tensor power [defining "infinite" here as "arbitrarily large and finite", since actual infinite-order tensor powers probably behave more strangely] of finite-dimensional vector spaces, of course disallowing infinite objects. The reason for this long-winded analogy is that it's basically a `SquareArray{T, ∞}`, which provides a baseline for what should and shouldn't work.) This is perfectly consistent algebraically but the algebra is unimportant. It's only a guide for what kind of functionality should be implemented.
- While the mathematics makes sense to me, I couldn't find any papers on treating Option values as infinite-order tensor powers of finite-dimensional vector spaces, and I wouldn't be surprised if this was not a description in use at all. So if this turns out to be absolutely incoherent, then let's abandon the analogy. I just thought that it is quite nice for explaining the parallel for `broadcast` and for indexing, as identical operations.

### DIFF
--- a/src/functional.jl
+++ b/src/functional.jl
@@ -5,107 +5,75 @@ module FunctionalNullableOperations
 importall Base
 import Base: promote_op, LinearFast
 
-# conceptually, nullable types can be considered infinite-order tensor powers of
-# finite-dimensional vector spaces — we attempt to support most operations on
-# arrays except the linear algebra related ones; the infinite-dimensional nature
-# makes subtyping AbstractArray a little dangerous, which explains why
-# functionality is reimplemented instead of subtyping AbstractArray
-
-# size   – not implemented since an infinite-dimensional tuple would be strange
-# length – one or zero
-# endof  – same as length
-length(u::Nullable) = u.isnull ? 0 : 1
-endof(u::Nullable)  = length(u)
+length(x::Nullable) = x.isnull ? 0 : 1
+endof(x::Nullable)  = length(x)
 
 # indexing is either without index, or with 1 as index
 # generalized linear indexing is not supported
 # setindex! not supported because Nullable is immutable
 linearindexing{T}(::Nullable{T}) = LinearFast()
-function getindex(u::Nullable)
-    @boundscheck u.isnull && throw(NullException())
-    u.value
+function getindex(x::Nullable)
+    @boundscheck x.isnull && throw(NullException())
+    x.value
 end
-function getindex(u::Nullable, i::Integer)
-    @boundscheck u.isnull | (i ≠ one(i)) && throw(BoundsError(i, u))
-    u.value
+function getindex(x::Nullable, i::Integer)
+    @boundscheck (x.isnull | (i ≠ 1)) && throw(BoundsError(i, x))
+    x.value
 end
 
 # iteration protocol
-start(u::Nullable) = 1
-next(u::Nullable, i::Integer) = u.value, 0
-done(u::Nullable, i::Integer) = u.isnull || i == 0
+start(x::Nullable) = 1
+next(x::Nullable, i::Integer) = x.value, 0
+done(x::Nullable, i::Integer) = x.isnull | (i == 0)
 
-# next we have reimplementations of some higher-order functions
-filter{T}(p, u::Nullable{T}) = u.isnull ? u : p(u.value) ? u : Nullable{T}()
+# higher-order functions
+function filter{T}(p, x::Nullable{T})
+    if x.isnull
+        x
+    elseif p(x.value)
+        x
+    else
+        Nullable{T}()
+    end
+end
+map{T}(f, x::Nullable{T}) = x.isnull ? Nullable{Union{}}() : Nullable(f(x.value))
 
-# warning: not type-stable
-map{T}(f, u::Nullable{T}) = u.isnull ? Nullable{Union{}}() : Nullable(f(u.value))
-
-# multi-argument map doesn't broadcast, so not very useful, but no harm having
-# it...
-function map(f, us::Nullable...)
-    if all(isnull, us)
+function map(f, xs::Nullable...)
+    if all(isnull, xs)
         Nullable()
-    elseif !any(isnull, us)
-        Nullable(map(f, map(getindex, us)...))
+    elseif !any(isnull, xs)
+        Nullable(map(f, map(getindex, xs)...))
     else
         throw(DimensionMismatch("expected all null or all nonnull"))
     end
 end
 
-# foldr and foldl are quite useful to express "do something if not null, else"
-# these
+broadcast{T}(f, x::Nullable{T}) =
+    x.isnull ? Nullable{promote_op(f, T)}() : Nullable{promote_op(f, T)}(f(x.value))
 
-# being infinite-dimensional, nullables are generally incompatible with
-# broadcast with arrays — it is probably not worth supporting the rare case
-# where an array has length 0 or 1
+# FIXME: need fast implementation for 2/3? args
 
-# so we reimplement broadcast, which has the same semantics but very different
-# implementation. This implementation is in fact much simpler than that for
-# arrays. Length-1 (non-nulls) are flexible and can broadcast to length-0
-# (nulls), but the other way does not work. Numbers are zero-dimensional and can
-# broadcast to infinite-dimensional nullables, but the other direction is not
-# supported.
-
-# Base's broadcast has a very loose signature so we can easily make it more
-# specific. Broadcast on numbers is still not supported. FIXME: remove generated
-# functions where unnecessary
-
-# some specialized functions
-broadcast{T}(f, u::Nullable{T}) =
-    u.isnull ? Nullable{promote_op(f, T)}() : Nullable{promote_op(f, T)}(f(u.value))
-
-@generated function broadcast(f, u::Union{Nullable,Number}, v::Union{Nullable,Number})
-    checkfor(s) = :($s.isnull && return Nullable{result}())
-    lifted = [T <: Nullable ? T.parameters[1] : T for T in (u, v)]
-    checks = vcat(u <: Nullable ? [checkfor(:u)] : [],
-                  v <: Nullable ? [checkfor(:v)] : [])
-    quote
-        result = promote_op(f, $(lifted...))
-        $(checks...)
-        @inbounds return Nullable{result}(f(u[], v[]))
-    end
-end
-
-# functions with three arguments or more are a bit expensive to specialize...
-# FIXME: why the arbitrary cutoff? justify
-function broadcast(f, us::Union{Nullable, Number}...)
+# generic multi-argument implementation
+function broadcast(f, xs::Union{Nullable, Number}...)
     result = promote_op(f,
-        [T <: Nullable ? T.parameters[1] : T for T in map(typeof, us)]...)
-    for u in us
-        if isa(u, Nullable) && u.isnull
+        [T <: Nullable ? T.parameters[1] : T for T in map(typeof, xs)]...)
+    for x in xs
+        if isa(x, Nullable) && x.isnull
             return Nullable{result}()
         end
     end
-    @inbounds return Nullable{result}(f(map(getindex, us)...))
+    @inbounds return Nullable{result}(f(map(getindex, xs)...))
 end
 
-# FIXME: these operations are probably not all correct
+# FIXME: these operations may be incorrect
 # and definitely some of them are slow, needs specialization
 # also have to be careful to avoid ambiguities... needs testing
-for eop in :(.+, .-, .*, ./, .\, .//, .==, .<, .!=, .<=, .÷, .%, .<<, .>>, .^).args
-    @eval $eop(u::Nullable, v::Union{Nullable, Number}) = broadcast($eop, u, v)
-    @eval $eop(u::Number, v::Nullable) = broadcast($eop, u, v)
+for (eop, op) in ((:.+, :+), (:.-, :-), (:.*, :*), (:./, :/), (:.\, :\),
+                  (:.//, ://), (:.==, :(==)), (:.<, :<), (:.!=, :!=),
+                  (:.<=, :<=), (:.÷, :÷), (:.%, :%), (:.<<, :<<), (:.>>, :>>),
+                  (:.^, :^))
+    @eval $eop(x::Nullable, y::Union{Nullable, Number}) = broadcast($op, x, y)
+    @eval $eop(x::Number, y::Nullable) = broadcast($op, x, y)
 end
 
 end # module
--- a/src/functional.jl
+++ b/src/functional.jl
@@ -67,21 +67,6 @@ end
 # broadcast to infinite-dimensional nullables, but the other direction is not
 # supported.
 
-# there are two shapes we are concerned about: infinite-dimensional 1×1×...
-# and infinite-dimensional 0×0×...; we don't care about zero-dimensional because
-# in that case all arguments were numbers, and broadcasting over only numbers
-# isn't supported by base currently
-function nullable_broadcast_shape(us::Union{Nullable,Number}...)
-    for u in us
-        if isa(us, Nullable)
-            if u.isnull
-                return true
-            end
-        end
-    end
-    return false
-end
-
 # Base's broadcast has a very loose signature so we can easily make it more
 # specific. Broadcast on numbers is still not supported. FIXME: remove generated
 # functions where unnecessary
--- a/src/functional.jl
+++ b/src/functional.jl
@@ -1,0 +1,126 @@
+module FunctionalNullableOperations
+
+# extends Base with operations that treat nullables as collections
+
+importall Base
+import Base: promote_op, LinearFast
+
+# conceptually, nullable types can be considered infinite-order tensor powers of
+# finite-dimensional vector spaces — we attempt to support most operations on
+# arrays except the linear algebra related ones; the infinite-dimensional nature
+# makes subtyping AbstractArray a little dangerous, which explains why
+# functionality is reimplemented instead of subtyping AbstractArray
+
+# size   – not implemented since an infinite-dimensional tuple would be strange
+# length – one or zero
+# endof  – same as length
+length(u::Nullable) = u.isnull ? 0 : 1
+endof(u::Nullable)  = length(u)
+
+# indexing is either without index, or with 1 as index
+# generalized linear indexing is not supported
+# setindex! not supported because Nullable is immutable
+linearindexing{T}(::Nullable{T}) = LinearFast()
+function getindex(u::Nullable)
+    @boundscheck u.isnull && throw(NullException())
+    u.value
+end
+function getindex(u::Nullable, i::Integer)
+    @boundscheck u.isnull | (i ≠ one(i)) && throw(BoundsError(i, u))
+    u.value
+end
+
+# iteration protocol
+start(u::Nullable) = 1
+next(u::Nullable, i::Integer) = u.value, 0
+done(u::Nullable, i::Integer) = u.isnull || i == 0
+
+# next we have reimplementations of some higher-order functions
+filter{T}(p, u::Nullable{T}) = u.isnull ? u : p(u.value) ? u : Nullable{T}()
+
+# warning: not type-stable
+map{T}(f, u::Nullable{T}) = u.isnull ? Nullable{Union{}}() : Nullable(f(u.value))
+
+# multi-argument map doesn't broadcast, so not very useful, but no harm having
+# it...
+function map(f, us::Nullable...)
+    if all(isnull, us)
+        Nullable()
+    elseif !any(isnull, us)
+        Nullable(map(f, map(getindex, us)...))
+    else
+        throw(DimensionMismatch("expected all null or all nonnull"))
+    end
+end
+
+# foldr and foldl are quite useful to express "do something if not null, else"
+# these
+
+# being infinite-dimensional, nullables are generally incompatible with
+# broadcast with arrays — it is probably not worth supporting the rare case
+# where an array has length 0 or 1
+
+# so we reimplement broadcast, which has the same semantics but very different
+# implementation. This implementation is in fact much simpler than that for
+# arrays. Length-1 (non-nulls) are flexible and can broadcast to length-0
+# (nulls), but the other way does not work. Numbers are zero-dimensional and can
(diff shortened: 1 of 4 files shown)


TotalVerb:
I have never seen such a large coverage decrease! [I turned the other tests off for faster testing locally, and forgot to turn them back on again.]

hayd:
It feels a little strange this being is in NullableArrays rather than base (or, if it really can't get into base/0.5, its own package)... 👍  to this functionality.

Edit: IMO The map, getindex are more obviously good for base, I suspect the broadcast stuff is distinct from that... and more controversial (?).

vchuravy:
Already map over Nullable is controversial in Base. https://github.com/JuliaLang/julia/pull/9446

src/functional.jl
```diff
@@ -0,0 +1,111 @@
+module FunctionalNullableOperations
+
+# extends Base with operations that treat nullables as collections
+
+importall Base
+import Base: promote_op, LinearFast
+
+# conceptually, nullable types can be considered infinite-order tensor powers of
```
nalimilan: I don't think this comment really helps understanding the code. Personally, I rather find it hard to get. :-)

src/functional.jl
```diff
@@ -0,0 +1,111 @@
...
+# finite-dimensional vector spaces — we attempt to support most operations on
+# arrays except the linear algebra related ones; the infinite-dimensional nature
+# makes subtyping AbstractArray a little dangerous, which explains why
+# functionality is reimplemented instead of subtyping AbstractArray
+
+# size   – not implemented since an infinite-dimensional tuple would be strange
```
nalimilan: Why not consider `Nullable` as a 0-dimensional array instead? It's really similar. Then its length could be the empty tuple:

``` julia
julia> x=Array(Int)
0-dimensional Array{Int64,0}:
139986151483248

julia> size(x)
()

```

src/functional.jl
```diff
@@ -0,0 +1,111 @@
...
+# length – one or zero
```
nalimilan: Please remove comments that repeat the code.

src/functional.jl
```diff
@@ -0,0 +1,111 @@
...
+# endof  – same as length
+length(u::Nullable) = u.isnull ? 0 : 1
```
nalimilan: Code in Base uses `x` instead of `u`, please stick to that convention.

src/functional.jl
```diff
@@ -0,0 +1,111 @@
...
+# indexing is either without index, or with 1 as index
+# generalized linear indexing is not supported
+# setindex! not supported because Nullable is immutable
+linearindexing{T}(::Nullable{T}) = LinearFast()
+function getindex(u::Nullable)
+    @boundscheck u.isnull && throw(NullException())
+    u.value
+end
+function getindex(u::Nullable, i::Integer)
+    @boundscheck u.isnull | (i ≠ one(i)) && throw(BoundsError(i, u))
```
nalimilan: I think `!= 1` should be enough. But you also need to check that `i` is `0` or `1`, all other values are incorrect.
Also, please add parentheses around the block before `&&` for clarity.

src/functional.jl
```diff
@@ -0,0 +1,111 @@
...
+    u.value
+end
+
+# iteration protocol
+start(u::Nullable) = 1
+next(u::Nullable, i::Integer) = u.value, 0
+done(u::Nullable, i::Integer) = u.isnull || i == 0
```
nalimilan: Maybe `|` here too. Not sure it makes a difference.

src/functional.jl
```diff
@@ -0,0 +1,111 @@
...
+
+# next we have reimplementations of some higher-order functions
```
nalimilan: Useless comment.

src/functional.jl
```diff
@@ -0,0 +1,111 @@
...
+filter{T}(p, u::Nullable{T}) = u.isnull ? u : p(u.value) ? u : Nullable{T}()
```
nalimilan: Could you use an `ifelse` for the outer condition to make this easier to read? Could potentially avoid a branch too.

src/functional.jl
```diff
@@ -0,0 +1,111 @@
...
+    if all(isnull, us)
+        Nullable()
+    elseif !any(isnull, us)
+        Nullable(map(f, map(getindex, us)...))
+    else
+        throw(DimensionMismatch("expected all null or all nonnull"))
+    end
+end
+
+# foldr and foldl are quite useful to express "do something if not null, else"
```
nalimilan: Again, this comment would better be kept in the PR discussion, and possibly in the manual (under a simpler form).

src/functional.jl
```diff
@@ -0,0 +1,111 @@
...
+            return Nullable{result}()
+        end
+    end
+    @inbounds return Nullable{result}(f(map(getindex, us)...))
+end
+
+# FIXME: these operations are probably not all correct
+# and definitely some of them are slow, needs specialization
+# also have to be careful to avoid ambiguities... needs testing
+for eop in :(.+, .-, .*, ./, .\, .//, .==, .<, .!=, .<=, .÷, .%, .<<, .>>, .^).args
```
nalimilan: Instead of `.args`, use symbols directly (i.e. `:.+`, etc.).

test/functional.jl
```diff
@@ -0,0 +1,108 @@
...
+
+isnullableof(u::Nullable, x) = !isnull(u) && u.value === x
+isnullableof(x) = u -> isnullableof(u, x)
+
+using Base.Test
+
+# getindex
+@test_throws NullException Nullable()[]
+@test_throws NullException Nullable{Int}()[]
+@test Nullable(0)[] == 0
```
nalimilan, line 18: Use `===` here.

nalimilan, line 18: (And same elsewhere.)

test/functional.jl
```diff
@@ -0,0 +1,108 @@
...
+
+@test_throws BoundsError Nullable()[1]
```
nalimilan, line 22: Also test a negative value, both on null and non-null. Test `Nullable()[2]`, and also `Nullable{Int}()`.

test/functional.jl
```diff
@@ -0,0 +1,108 @@
...
+@test_throws BoundsError Nullable(0)[0]
+@test_throws BoundsError Nullable(0)[2]
+@test Nullable(0)[1] == 0
+
+# collect
+@test collect(Nullable()) == Union{}[]
```
nalimilan: Be careful, `==` doesn't test the type at all. You need a separate test using `typeof`.

test/functional.jl
```diff
@@ -0,0 +1,108 @@
...
+@test collect(Nullable{Int}()) == Int[]
+@test collect(Nullable(85)) == Int[85]
+@test collect(Nullable(1.0)) == Float64[1.0]
+
+# length
```
nalimilan: And `endof`. Also test `endof` on a null.

test/functional.jl
```diff
@@ -0,0 +1,108 @@
...
+@test length(Nullable()) == 0
+@test length(Nullable{Int}()) == 0
+@test length(Nullable(1.0)) == 1
+@test endof(Nullable(85)) == 1
+
+# filter
+for p in (_ -> true, _ -> false)
+    @test filter(p, Nullable())      |> isnull_typed(Union{})
```
nalimilan, line 45: I don't see the point of this `isnull_typed` function. Why not do `=== Nullable{Union{}}()`?

test/functional.jl
```diff
@@ -0,0 +1,108 @@
...
+@test map(sqr, Nullable(2))     |> isnullableof(4)
+
+@test map(+, Nullable(1), Nullable(2), Nullable(3)) |> isnullableof(6)
+@test map(+, Nullable(), Nullable(), Nullable())    |> isnull_typed(Union{})
+@test map(+, Nullable{Int}(), Nullable{Int}())      |> isnull_typed(Union{})
+
+# example: square if value exists, -1 if value is null
+# with foldl/foldr/reduce
+for fn in (foldl, foldr, reduce)
+    @test foldl((_, x) -> x^2, -1, Nullable())   == -1
```
nalimilan: You forgot to use `fn` here.

test/functional.jl
```diff
@@ -0,0 +1,108 @@
...
+    @test foldl((_, x) -> x^2, -1, Nullable(10)) == 100
+end
+
+# with broadcast and get (map does not work because of get limitations...)
```
nalimilan: What do you mean?

test/functional.jl
```diff
@@ -0,0 +1,108 @@
...
+# perhaps the get limitations should be fixed
+@test get(sqr.(Nullable{Int}()), -1) == -1
```
nalimilan: You're testing `get` as much as `map` with this. Do `sqr.(Nullable{Int}()) === Nullable{Union{}}()` instead.

test/functional.jl
```diff
@@ -0,0 +1,108 @@
...
+# broadcast and elementwise
+@test sin.(Nullable(0.0))             |> isnullableof(0.0)
+@test sin.(Nullable{Float64}())       |> isnull_typed(Float64)
+
+@test Nullable(8) .+ Nullable(10)     |> isnullableof(18)
+@test Nullable(8) .- Nullable(10)     |> isnullableof(-2)
+@test Nullable(8) .+ Nullable{Int}()  |> isnull_typed(Int)
+@test Nullable{Int}() .- Nullable(10) |> isnull_typed(Int)
+
+@test log.(10, Nullable(1.0))         |> isnullableof(0.0)
```
nalimilan: Could you also test with `Nullable(10)` as the first argument?

test/functional.jl
```diff
@@ -0,0 +1,108 @@
...
+@test log.(10, Nullable{Float64}())   |> isnull_typed(Float64)
+
+@test Nullable(2) .^ Nullable(4)      |> isnullableof(16)
+@test Nullable(2) .^ Nullable{Int}()  |> isnull_typed(Int)
+
+# big broadcast (slow)
```
nalimilan: "Slow" isn't relevant inside test code. Please remove this mention as well as the one below.

src/functional.jl
```diff
@@ -0,0 +1,111 @@
...
+next(u::Nullable, i::Integer) = u.value, 0
+done(u::Nullable, i::Integer) = u.isnull || i == 0
+
+# next we have reimplementations of some higher-order functions
+filter{T}(p, u::Nullable{T}) = u.isnull ? u : p(u.value) ? u : Nullable{T}()
+
+# warning: not type-stable
+map{T}(f, u::Nullable{T}) = u.isnull ? Nullable{Union{}}() : Nullable(f(u.value))
+
+# multi-argument map doesn't broadcast, so not very useful, but no harm having
```
nalimilan: Idem.

src/functional.jl
```diff
@@ -0,0 +1,111 @@
...
+# it...
+function map(f, us::Nullable...)
+    if all(isnull, us)
+        Nullable()
+    elseif !any(isnull, us)
+        Nullable(map(f, map(getindex, us)...))
+    else
+        throw(DimensionMismatch("expected all null or all nonnull"))
```
nalimilan, line 47: Why did you choose this behavior? For mathematic logic? Wouldn't it be more useful to return null if one of the arguments is null?

src/functional.jl
```diff
@@ -0,0 +1,111 @@
...
+# specific. Broadcast on numbers is still not supported. FIXME: remove generated
+# functions where unnecessary
+
+# some specialized functions
+broadcast{T}(f, u::Nullable{T}) =
+    u.isnull ? Nullable{promote_op(f, T)}() : Nullable{promote_op(f, T)}(f(u.value))
+
+@generated function broadcast(f, u::Union{Nullable,Number}, v::Union{Nullable,Number})
+    checkfor(s) = :($s.isnull && return Nullable{result}())
+    lifted = [T <: Nullable ? T.parameters[1] : T for T in (u, v)]
```
nalimilan: `T.parameters` can be replaced with `eltype(T)`. Also, I would find it clearer to compute `U` and `V` separately instead of creating a vector and splatting it below.

src/functional.jl
```diff
@@ -0,0 +1,111 @@
...
+    checks = vcat(u <: Nullable ? [checkfor(:u)] : [],
+                  v <: Nullable ? [checkfor(:v)] : [])
+    quote
+        result = promote_op(f, $(lifted...))
+        $(checks...)
+        @inbounds return Nullable{result}(f(u[], v[]))
```
nalimilan: Access `u.value` and `v.value` directly instead of `@inbounds`.

src/functional.jl
```diff
@@ -0,0 +1,111 @@
...
+
+# Base's broadcast has a very loose signature so we can easily make it more
+# specific. Broadcast on numbers is still not supported. FIXME: remove generated
+# functions where unnecessary
+
+# some specialized functions
+broadcast{T}(f, u::Nullable{T}) =
+    u.isnull ? Nullable{promote_op(f, T)}() : Nullable{promote_op(f, T)}(f(u.value))
+
+@generated function broadcast(f, u::Union{Nullable,Number}, v::Union{Nullable,Number})
```
nalimilan: You really don't need a generated function here AFAICT. Just define helper methods like:

``` julia
typeof_lifted{T}(x::T) = T
typeof_lifted{T}(x::Nullable{T}) = T
isnull_anytype(x::Any) = false
isnull_anytype(x::Nullable) = isnull(x)
```

The latter is so commonly needed that I wonder whether we should export something like this from Base (e.g.`isnull(x, false)` could return `fase` on non-`Nullable`).

nalimilan:
I agree this should go into Base (`base/nullable.jl`), not NullableArrays. We're rather trying to move other essential features into Base.

There are many points to discuss in the description, but let's try to keep focused on this PR. I think we should keep `get` for now, I don't really see the problem with it: we could even add `get(x::AbstractArray, i)` (with no default) for consistency, even if it would be redundent with `getindex`. Anyway that's a minor point which could be tackled later. The `get(Nullable(), 1)` failure sounds like a bug to file separately. As regards `==`, see the arithmetic operators issues I linked to at https://github.com/JuliaLang/julia/issues/16889#issuecomment-225418027.

Regarding type-inference issues, they are dependent on fixing `broadcast` and `map` in Base. See e.g. https://github.com/JuliaLang/julia/issues/4883.

@vchuravy I think with the new `.(` syntax it's less controversial. At least I now find this approach interesting.

TotalVerb:
Thanks for the comments. I see that NullableArrays is not the right place to submit this PR. After addressing @nalimilan's comments [thanks!], I'll move this PR to Base.

test/functional.jl
TotalVerb: this was the `get(::Nullable{Union{}})` issue I mentioned and you submitted a fix for. Since map returns `Nullable{Union{}}`, `get`ting that with a default value fails.

test/functional.jl
TotalVerb, line 45: yes, that would work. I wasn't sure if `===` was correct for bools in Julia (in many languages booleans can take values other than 0 or 1).

src/functional.jl
TotalVerb: I'm not so attached either. I'll leave this comment out.

src/functional.jl
TotalVerb: Unfortunately, this is inaccurate. 0-dimensional arrays cannot contain 0 elements.

src/functional.jl
TotalVerb: Sounds good.

src/functional.jl
TotalVerb: OK.

src/functional.jl
TotalVerb: i must be 1, and this is checked already. Added parentheses.

src/functional.jl
TotalVerb: I don't think it does, but consistency is good.

src/functional.jl
TotalVerb: Would `# higher-order functions` be better?

src/functional.jl
TotalVerb: OK.

src/functional.jl
TotalVerb: Sorry about the sloppiness! I used these comments as a bit of a guide for implementation. I should take them out.

src/functional.jl
TotalVerb: That was just me being lazy. Fixed.

test/functional.jl
TotalVerb, line 18: :+1:

test/functional.jl
TotalVerb, line 22: Very reasonable. Done.

test/functional.jl
TotalVerb: Good catch. Used `eltype`.

test/functional.jl
TotalVerb, line 45: while `=== Nullable{Union{}}()` works, the same for `Int` doesn't. I'll keep the `isnull_typed` function for now.

test/functional.jl
TotalVerb: oops

test/functional.jl
TotalVerb: These tests are the same as the ones below. I'll leave them out.

test/functional.jl
TotalVerb: Good idea.

test/functional.jl
TotalVerb: :+1:

src/functional.jl
TotalVerb: left out the comments

src/functional.jl
TotalVerb, line 47: that's not what map does on arrays. I don't like magic autolifting, especially since there's a very similar function that has the lifting behaviour.

src/functional.jl
TotalVerb: good comments here (and below). I'll try rewriting the implementation of short broadcast.

src/functional.jl
TotalVerb: I actually can't eliminate the branch in general, since `x.value` isn't defined for certain types (`Union{}` in particular). `NullableArrays` uses an `isbits` test for this. Once I optimize for performance, I'll try out a similar trick.

test/functional.jl
TotalVerb: I removed `reduce` here because the operator we're testing is non-associative. `foldr` uses the opposite argument order so we need separate tests for that.

TotalVerb:
Thank you @nalimilan for the review comments! Much appreciated. I will close this PR after Travis passes and move it to Base later tomorrow.

TotalVerb:
Travis passing now (except for unrelated, as far as I can tell, failure on nightly). Will move PR to Base.

src/functional.jl
nalimilan: OK, let's leave `size` unimplemented. Anyway it's probably not very useful.

src/functional.jl
nalimilan: Indeed, I misread the condition. Regarding parentheses and `!= 1`, I meant this:

``` julia
@boundscheck (u.isnull | i != 1) && throw(BoundsError(i, u))
```

src/functional.jl
nalimilan: Why not. Depends on how `base/nullable.jl` is organised.

src/functional.jl
nalimilan: Good catch. Specializing depending on `isbits` makes sense.

test/functional.jl
nalimilan, line 45: Yeah, that's https://github.com/JuliaLang/julia/issues/16923.

Issue #116, opened 2016-06-15
Some nullable operations can throw exceptions


TotalVerb:
Not all of the operations defined [here](https://github.com/JuliaStats/NullableArrays.jl/blob/master/src/operators.jl) are safe, because some operations are not defined on the entire domain, even if the types are bitstypes. Null values should not cause exceptions, especially in unpredictable ways.

Example:

```
julia> Nullable{Int}() ^ (Nullable{Int}() - Nullable{Int}() - Nullable{Int}())
ERROR: DomainError:
Cannot raise an integer x to a negative power -n. 
Make x a float by adding a zero decimal (e.g. 2.0^-n instead of 2^-n), or write 1/x^n, float(x)^-n, or (x//1)^-n.
 in power_by_squaring(::Int64, ::Int64) at ./intfuncs.jl:85
 in ^(::Nullable{Int64}, ::Nullable{Int64}) at /home/fengyang/.julia/v0.5/NullableArrays/src/operators.jl:38
 in eval(::Module, ::Any) at ./boot.jl:225
 in macro expansion at ./REPL.jl:92 [inlined]
 in (::Base.REPL.##1#2{Base.REPL.REPLBackend})() at ./event.jl:46
```

(this happens to crash on my REPL fairly often, but your mileage may vary)

It's somewhat rare to actually have issues, but the following three operations should definitely be removed:
- `%` (doesn't work with 0)
- `^` (doesn't work with negative exponents)
- `sqrt` (doesn't work with negatives)

nalimilan:
Actually, no operation is really safe for all possible types, even bitstypes. For example, a checked integer type would throw an error in case of overflow with `+`. I think what we need is a an internal `undef_safe_op(f::Function, T::Type...)::Bool` which we would define as `true` for standard bitstypes and standard operations which are OK to call on any value (i.e. not the ones you list above).

Since you seem to be willing to work on Nullable, would you want to open a PR against Base to implement the operators which currently live in this package, but in fully safe approach? Cf. https://github.com/JuliaStats/NullableArrays.jl/issues/111.

davidagold:
Would a more appropriate name be `null_safe_op`? Also, how would such a solution solve the overflow problem? That seems to be distinct from (yet related to) what @TotalVerb cites in their original post.

nalimilan:
[...]

Issue #111, opened 2016-05-28
Arithmetic operators on Nullable


bkamins:
The line:
https://github.com/JuliaStats/NullableArrays.jl/blob/master/src/operators.jl#L37
has the following consequences:

a) types that are not bits throw error when operated on:

```
x = Nullable(1)
x+x # works
x = Nullable(BigInt(1))
x+x # error
```

This probably should be fixed.

b) operation on a missing of type that is not bits fails

```
x = Nullable(1.0)
y = Nullable()
x+y # error
x+convert(Nullable{Int64}, y) # works as addition of Int64 and Float64 is possible
```

I am not sure if this should be fixed. Implementation `convert` for `Nullable` implies that missing value of `Nullable{T}` can be always converted to `Nullable{S}` independent on what `T` and `S` are but I am not sure if this conversion should be automatically performed in operations on nullables.

nalimilan:
Thanks for looking into this. We should fix it in Julia Base as part of https://github.com/JuliaLang/julia/pull/16080.

I guess the `isbits` restriction was made to get rid of a branch, as it allows using `+` even when the value is missing. But we should definitely support other types as well. Automatic promotion should also be supported.

johnmyleswhite:
This kind of change needs to be made with some caution. As is, the code is already unsafe since we're implicitly assuming that it's always safe to operate on the raw bits associated with a value: you can create nonsensical immutables this way and it's not impossible to imagine scenarios in which our current strategy introduces bugs. We will almost surely introduce segfaults if we remove this restriction without substantial code changes -- but those code changes could prohibit SIMD optimizations, so care will be required.

The current approach is a way of working around the lack of return types. Making these things both fast and safe requires that we know what type of nullable we should return even when the arguments are nulls.

nalimilan:
> This kind of change needs to be made with some caution. As is, the code is already unsafe since we're implicitly assuming that it's always safe to operate on the raw bits associated with a value: you can create nonsensical immutables this way and it's not impossible to imagine scenarios in which our current strategy introduces bugs. We will almost surely introduce segfaults if we remove this restriction without substantial code changes -- but those code changes could prohibit SIMD optimizations, so care will be required.

You mean that adding anything to the `else` branch would break SIMD? I wonder whether LLVM shouldn't be able to perform loop unswitching here.

> The current approach is a way of working around the lack of return types. Making these things both fast and safe requires that we know what type of nullable we should return even when the arguments are nulls.

[...]

PR #119, opened 2016-06-18
Improve operators

nalimilan:
Introduce a new `null_safe_op()` function which is used instead of
`isbits()` to decide whether to take the fast path (without a branch)
or the slow path in lifted operators. This is more correct (and thus
safer), and allows supporting operations on any type. Also add an
extensive set of tests for both paths.

On Julia 0.4, operators for which no functor exist cannot be identified
by `null_safe_op()`: for these, fall back to the slow path on that version.

This PR introduces  `cbrt` as a new supported operation since it has a special
operator `∛`. It also keeps `abs()`, `abs2()`, `scalarmin()` and `scalarmax()`, which
should probably be deprecated once we have a more general solution for
lifting.

Tests pass on Julia 0.4, but on 0.5 they need https://github.com/JuliaLang/julia/pull/16995 (unrelated tests already fail too).

Fixes https://github.com/JuliaStats/NullableArrays.jl/issues/74, https://github.com/JuliaStats/NullableArrays.jl/issues/111, https://github.com/JuliaStats/NullableArrays.jl/issues/116.

### DIFF
--- a/src/operators.jl
+++ b/src/operators.jl
@@ -1,121 +1,128 @@
-import Compat: @functorize
+## Lifted operators
 
-@noinline throw_error() = error()
+importall Base.Operators
+import Base: promote_op, abs, abs2, sqrt, cbrt, scalarmin, scalarmax
+using Compat: @functorize
 
-for f in (
-    :(@compat Base.:+),
-    :(@compat Base.:-),
-    :(@compat Base.:!),
-    :(@compat Base.:~),
-)
+# Methods adapted from Base Julia 0.5
+if VERSION <= v"0.5.0-dev"
+    promote_op(::Any, T) = T
+    promote_op{T}(::Type{T}, ::Any) = T
+
+    promote_op{R<:Number}(op, ::Type{R}) = typeof(op(one(R)))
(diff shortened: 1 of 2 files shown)


src/operators.jl
```diff
@@ -1,121 +1,130 @@
...
-    :(@compat Base.:|),
-    :(@compat Base.:^),
-    :(@compat Base.:<<),
-    :(@compat Base.:>>),
-)
+"""
+    null_safe_op(f::Any, ::Type)::Bool
+    null_safe_op(f::Any, ::Type, ::Type)::Bool
+
+Returns whether an operation `f` can safely be applied to any value of the passed type(s).
```
johnmyleswhite, line 20: Minor nit: maybe we should distinguish between value and bit pattern? Consider the following type:

```
immutable OneType
    x::Int

    function OneType()
        new(1)
    end
end
```

The only legal value of this type is `1`, but our uninitialized constructors will generate bit patterns that are not legal values.

nalimilan, line 20: Maybe we should mention "bit pattern" below instead of "domain"? I think "value" is good as a first approximation to give the idea of the requirements.

nalimilan, line 20: OK, I've updated the docstring.

nalimilan:
With support merged in Julia, tests now pass. Will tag a new release.

PR #128, opened 2016-07-06
map revisions

### LINKED ISSUE #130: map is broken on v0.0.5
This will be addressed by #128 , but it needs to be noted:

```
               _
   _       _ _(_)_     |  A fresh approach to technical computing
  (_)     | (_) (_)    |  Documentation: http://docs.julialang.org
   _ _   _| |_  __ _   |  Type "?help" for help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 0.5.0-dev+5156 (2016-07-05 13:54 UTC)
 _/ |\__'_|_|_|\__'_|  |  Commit 0b22a01* (5 days old master)
|__/                   |  x86_64-apple-darwin15.5.0

julia> using NullableArrays

julia> map(sin, NullableArray(Int[1]); lift=true)
ERROR: UndefVarError: f not defined
 in (::NullableArrays.##_F_#35#39)(::Bool, ::Function, ::NullableArrays.NullableArray{Int64,1}) at /Users/David/.julia/v0.5/NullableArrays/src/map.jl:90
 in (::NullableArrays.#kw##_F_#38)(::Array{Any,1}, ::NullableArrays.#_F_#38, ::NullableArrays.NullableArray{Int64,1}) at ./<missing>:0
 in (::NullableArrays.##map#7#8{Dict{Int64,Dict{Union{DataType,Function},Function}}})(::Bool, ::Function, ::Base.#sin, ::NullableArrays.NullableArray{Int64,1}, ::Vararg{NullableArrays.NullableArray{Int64,1},N}) at /Users/David/.julia/v0.5/NullableArrays/src/map.jl:172
 in (::Base.#kw##map)(::Array{Any,1}, ::Base.#map, ::Base.#sin, ::NullableArrays.NullableArray{Int64,1}) at ./<missing>:0
 in eval(::Module, ::Any) at ./boot.jl:234
 in macro expansion at ./REPL.jl:92 [inlined]
 in (::Base.REPL.##1#2{Base.REPL.REPLBackend})() at ./event.jl:46
```

davidagold:
This is an attempt to transition away from the metaprogramming-heavy approach to map we took for Julia v0.4. Admittedly, this does introduce the use of generated functions in n-arg `map`/`map!`. ~~Does anybody have any suggestions as to how best maintain compatibility with 0.4? What I have here seems kind of crude.~~

There's a lot of code repetition due to the 1-arg, 2-arg and n-arg methods for `map`. The Base `map(f, A)` regime calls `collect_similar(A, Generator(f, A))`, but I haven't found a way to `collect` over a `Generator{NullableArray, F}` without creating a lot of unnecessary `Nullable` objects -- so, hooking into that interface isn't really an option. I'm not sure the best way to reduce the code footprint here.

Here are the significant points for this PR:
- fix #130
- remove as much dependence on metaprogramming as possible
- Settle on well-defined behavior for the two "emtpy" cases: (i) `map` over empty `NullableArray` and (ii) `map over non-empty but all-null`NullableArray`. 

For both cases (i) and (ii) above, the call to `map` can be lifted or not. If `map` is not lifted, it falls back on the implementation in Base, which differs between 0.4 and 0.5 (and also between the 1-arg, 2-arg and n-arg implementations). On 0.5, we have the following four combinations of behaviors:

1) `map(f, X; lift=false)`, where `X` is empty. On 0.5, now that https://github.com/JuliaLang/julia/pull/16622 has been merged, this uses type inference (`Base.return_types`) to select the type parameter for the returned empty `NullableArray{T}`. Note that, since `f` is not lifted, `return_types` is called on `f, (eltype(X),)`. If somewhere down the line a method specialized on `Nullable` arguments is missing, this will return an empty `NullableArray{Union{}}`:

``` julia
julia> X = NullableArray(Int[])
0-element NullableArrays.NullableArray{Int64,1}

julia> f(x) = 5*x
f (generic function with 1 method)

julia> map(f, X)
0-element NullableArrays.NullableArray{Union{},1}

julia> h(x::Nullable) = Nullable(5*x.value, x.isnull)
h (generic function with 1 method)

julia> map(h, X)
0-element NullableArrays.NullableArray{Int64,1}
```

2) `map(f, X; lift=true)`, where `X` is empty. This uses type inference (`Base.return_types`) to select the type parameter for the returned empty `NullableArray`. Since `f` is lifted, `return_types` is called on `f, (inner_eltype(X),)`, where `inner_eltype(X::NullableArray{T}) = T`:

``` julia
julia> map(f, X; lift=true)
0-element NullableArrays.NullableArray{Int64,1}
```

3) `map(f, X; lift=false)` where `X` is non-empty but all null. The returned `NullableArray{T}` does _not_ depend on type inference --- it depends entirely on the behavior of `f` applied to empty `Nullable{T}` objects.

``` julia
julia> map(f, Y)
ERROR: MethodError: no method matching *(::Int64, ::Nullable{Int64})
Closest candidates are:
  *(::Any, ::Any, ::Any, ::Any...)
  *{S}(::Nullable{Union{}}, ::Nullable{S})
  *{T<:Union{Int128,Int16,Int32,Int64,Int8,UInt128,UInt16,UInt32,UInt64,UInt8}}(::T<:Union{Int128,Int16,Int32,Int64,Int8,UInt128,UInt16,UInt32,UInt64,UInt8}, ::T<:Union{Int128,Int16,Int32,Int64,Int8,UInt128,UInt16,UInt32,UInt64,UInt8})
  ...
 in _collect(::NullableArrays.NullableArray{Int64,1}, ::Base.Generator{NullableArrays.NullableArray{Int64,1},#f}, ::Base.EltypeUnknown, ::Base.HasShape) at ./array.jl:283
 in #map#11 at /Users/David/.julia/v0.5/NullableArrays/src/map.jl:108 [inlined]
 in map(::Function, ::NullableArrays.NullableArray{Int64,1}) at /Users/David/.julia/v0.5/NullableArrays/src/map.jl:108
 in eval(::Module, ::Any) at ./boot.jl:234
 in macro expansion at ./REPL.jl:92 [inlined]
 in (::Base.REPL.##1#2{Base.REPL.REPLBackend})() at ./event.jl:46

julia> map(h, Y)
3-element NullableArrays.NullableArray{Int64,1}:
 #NULL
 #NULL
 #NULL
```

4) `map(f, X; lift=true)` where `X` is non-empty but all null. This _does_ use type inference to determine the return type of the returned `NullableArray`. 

``` julia
julia> map(f, Y; lift=true)
3-element NullableArrays.NullableArray{Int64,1}:
 #NULL
 #NULL
 #NULL
```

Brief performance summary (on 0.5) (NOTE: revised as of 7/12/16):

``` julia
using NullableArrays
using BenchmarkTools
n = 1_000_000
A = rand(n)
M = rand(Bool, n)
X = NullableArray(A)
Y = NullableArray(A, M)
As = [ rand(n) for i in 1:5 ]
Ms = [ rand(Bool, n) for i in 1:5 ]
Xs = [ NullableArray(A) for A in As ]
Ys = [ NullableArray(A, M) for (A, M) in zip(As, Ms)]
f(x) = 5*x
f(x::Nullable) = Nullable(5)*x
f(xs...) = prod(xs)

julia> @benchmark map(f, A)
BenchmarkTools.Trial: 
  samples:          1094
  evals/sample:     1
  time tolerance:   5.00%
  memory tolerance: 1.00%
  memory estimate:  7.63 mb
  allocs estimate:  4
  minimum time:     2.79 ms (0.00% GC)
  median time:      3.38 ms (0.00% GC)
  mean time:        4.57 ms (21.49% GC)
  maximum time:     21.65 ms (0.00% GC)

julia> @benchmark map(f, X)
BenchmarkTools.Trial: 
  samples:          400
  evals/sample:     1
  time tolerance:   5.00%
  memory tolerance: 1.00%
  memory estimate:  8.58 mb
  allocs estimate:  12
  minimum time:     9.47 ms (0.00% GC)
  median time:      12.01 ms (0.00% GC)
  mean time:        12.51 ms (8.02% GC)
  maximum time:     36.04 ms (0.00% GC)

julia> @benchmark map(f, X; lift=true)
BenchmarkTools.Trial: 
  samples:          895
  evals/sample:     1
  time tolerance:   5.00%
  memory tolerance: 1.00%
  memory estimate:  8.58 mb
  allocs estimate:  9
  minimum time:     3.83 ms (0.00% GC)
  median time:      4.59 ms (0.00% GC)
  mean time:        5.58 ms (16.47% GC)
  maximum time:     65.67 ms (0.00% GC)

julia> @benchmark map(f, Y)
BenchmarkTools.Trial: 
  samples:          298
  evals/sample:     1
  time tolerance:   5.00%
  memory tolerance: 1.00%
  memory estimate:  8.58 mb
  allocs estimate:  12
  minimum time:     14.41 ms (0.00% GC)
  median time:      16.95 ms (0.00% GC)
  mean time:        16.82 ms (5.38% GC)
  maximum time:     21.59 ms (18.36% GC)

julia> @benchmark map(f, Y; lift=true)
BenchmarkTools.Trial: 
  samples:          582
  evals/sample:     1
  time tolerance:   5.00%
  memory tolerance: 1.00%
  memory estimate:  8.58 mb
  allocs estimate:  9
  minimum time:     6.05 ms (0.00% GC)
  median time:      7.84 ms (0.00% GC)
  mean time:        8.60 ms (12.49% GC)
  maximum time:     28.21 ms (44.18% GC)

julia> @benchmark map(f, As...)
BenchmarkTools.Trial: 
  samples:          222
  evals/sample:     1
  time tolerance:   5.00%
  memory tolerance: 1.00%
  memory estimate:  7.63 mb
  allocs estimate:  9
  minimum time:     19.83 ms (0.00% GC)
  median time:      22.67 ms (0.00% GC)
  mean time:        22.59 ms (4.04% GC)
  maximum time:     30.93 ms (8.61% GC)

@benchmark map(f, Xs...)
BenchmarkTools.Trial: 
  samples:          2
  evals/sample:     1
  time tolerance:   5.00%
  memory tolerance: 1.00%
  memory estimate:  2.10 gb
  allocs estimate:  61995904
  minimum time:     4.87 s (9.77% GC)
  median time:      4.97 s (9.63% GC)
  mean time:        4.97 s (9.63% GC)
  maximum time:     5.07 s (9.49% GC)

julia> @benchmark map(f, Xs...; lift=true)
BenchmarkTools.Trial: 
  samples:          291
  evals/sample:     1
  time tolerance:   5.00%
  memory tolerance: 1.00%
  memory estimate:  8.58 mb
  allocs estimate:  15
  minimum time:     12.57 ms (0.00% GC)
  median time:      15.94 ms (0.00% GC)
  mean time:        17.18 ms (5.67% GC)
  maximum time:     80.23 ms (0.00% GC)

julia> @benchmark map(f, Ys...)
BenchmarkTools.Trial: 
  samples:          1
  evals/sample:     1
  time tolerance:   5.00%
  memory tolerance: 1.00%
  memory estimate:  2.10 gb
  allocs estimate:  61995904
  minimum time:     5.30 s (9.30% GC)
  median time:      5.30 s (9.30% GC)
  mean time:        5.30 s (9.30% GC)
  maximum time:     5.30 s (9.30% GC)

@benchmark map(f, Ys...; lift=true)
BenchmarkTools.Trial: 
  samples:          350
  evals/sample:     1
  time tolerance:   5.00%
  memory tolerance: 1.00%
  memory estimate:  8.58 mb
  allocs estimate:  15
  minimum time:     12.21 ms (0.00% GC)
  median time:      14.08 ms (0.00% GC)
  mean time:        14.31 ms (6.22% GC)
  maximum time:     22.85 ms (11.79% GC)
```

~~It seems pretty clear that, as far as performance goes, we ought to transition towards lifting by default. I think this seems reasonable for most statistics/data science-related use cases. To do so, would we need to deprecate the current behavior for a release cycle?~~ [7/12/16]: Lifting will probably yield best performance, but clearly something is sub-optimal with the (n-arg especially) non-lifted `map` implementation.  

@nalimilan @johnmyleswhite thoughts?

### DIFF
--- a/src/map.jl
+++ b/src/map.jl
@@ -1,174 +1,214 @@
-using Base.Cartesian
+if VERSION < v"0.5.0-dev+3294"
+    include("map0_4.jl")
+else
+    using Base: collect_similar, Generator, ith_all
 
-function gen_nullcheck(narrays::Int)
-    As = [Symbol("A_"*string(i)) for i = 1:narrays]
-    e_nullcheck = :($(As[1]).isnull[i])
-    for i = 2:narrays
-        e_nullcheck = Expr(:||, e_nullcheck, :($(As[i]).isnull[i]))
+    macro nullcheck(Xs, nargs)
+        res = :($(Xs)[1].isnull[i])
+        for i = 2:nargs
+            e = :($(Xs)[$i].isnull[i])
+            res = Expr(:||, res, e)
+        end
+        return res
     end
-    return e_nullcheck
-end
 
-function gen_map!_body{F}(narrays::Int, lift::Bool, f::F)
-    _f = Expr(:quote, f)
-    e_nullcheck = gen_nullcheck(narrays)
-    if lift
-        return quote
-            for i in 1:length(dest)
-                if $e_nullcheck
+    macro fcall(Xs, nargs)
+        res = Expr(:call, :f)
+        for i in 1:nargs
+            push!(res.args, :($(Xs)[$i].values[i]))
+        end
+        return res
+    end
+
+    # Base.map!
+
+    Base.map!{F}(f::F, X::NullableArray; lift=false) = map!(f, X, X; lift=lift)
+    function Base.map!{F}(f::F, dest::NullableArray, X::NullableArray; lift=false)
+        if lift
+            for (i, j) in zip(eachindex(dest), eachindex(X))
+                if X.isnull[j]
                     dest.isnull[i] = true
                 else
-                    dest[i] = $_f((@ntuple $narrays j->A_j.values[i])...)
+                    dest.isnull[i] = false
+                    dest.values[i] = f(X.values[j])
                 end
             end
-        end
-    else
-        return quote
-            for i in 1:length(dest)
-                dest[i] = $_f((@ntuple $narrays j->A_j[i])...)
+        else
+            for (i, j) in zip(eachindex(dest), eachindex(X))
+                dest[i] = f(X[j])
             end
         end
+        return dest
     end
-end
 
-function gen_map_to!_body{F}(_map_to!::Symbol, narrays::Int, f::F)
-    _f = Expr(:quote, f)
-    e_nullcheck = gen_nullcheck(narrays)
-    return quote
-        @inbounds for i in offs:length(A_1)
-            if lift
-                if $e_nullcheck
-                    # we don't need to compute anything if A.isnull[i], since
-                    # the return type is specified by T and by the algorithm in
-                    # 'body'
+    function Base.map!{F}(f::F, dest::NullableArray, X1::NullableArray,
+                          X2::NullableArray; lift=false)
+        if lift
+            for (i, j, k) in zip(eachindex(dest), eachindex(X1), eachindex(X2))
+                if X1.isnull[j] | X2.isnull[k]
                     dest.isnull[i] = true
-                    continue
                 else
-                    v = $_f((@ntuple $narrays j->A_j.values[i])...)
+                    dest.isnull[i] = false
+                    dest.values[i] = f(X1.values[j], X2.values[k])
+                end
+            end
+        else
+            for (i, j, k) in zip(eachindex(dest), eachindex(X1), eachindex(X2))
+                dest[i] = f(X1[j], X2[k])
+            end
+        end
+        return dest
+    end
+
+    function Base.map!{F}(f::F, dest::NullableArray, Xs::NullableArray...; lift=false)
+        _map!(f, dest, Xs, lift)
+    end
+
+    @generated function _map!{F, N}(f::F, dest::NullableArray, Xs::Tuple{Vararg{NullableArray, N}}, lift)
+        return quote
+            if lift
+                for i in linearindices(Xs[1])
+                    if @nullcheck Xs $N
+                        dest.isnull[i] = true
+                    else
+                        dest.isnull[i] = false
+                        dest.values[i] = @fcall Xs $N
+                    end
                 end
             else
-                v = $_f((@ntuple $narrays j->A_j[i])...)
+                for i in linearindices(Xs[1])
+                    dest[i] = f(ith_all(i, Xs)...)
+                end
             end
-            S = typeof(v)
-            if S !== T && !(S <: T)
+            return dest
+        end
+    end
+
+    # Base.map
+
+    function Base.map(f, X::NullableArray; lift=false)
+        lift ? _liftedmap(f, X) : collect_similar(X, Generator(f, X))
+    end
+    function Base.map(f, X1::NullableArray, X2::NullableArray; lift=false)
+        lift ? _liftedmap(f, X1, X2) : collect(Generator(f, X1, X2))
+    end
+    function Base.map(f, Xs::NullableArray...; lift=false)
+        lift ? _liftedmap(f, Xs...) : collect(Generator(f, Xs...))
+    end
+
+    function _liftedmap(f, X::NullableArray)
+        len = length(X)
+        # if X is empty, fall back on type inference
+        len > 0 || return NullableArray{Base.return_types(f, (eltype(X),)), 0}()
+        i = 1
+        while X.isnull[i]
+            i += 1
+        end
+        # if X is all null, fall back on type inference
+        i <= len || return similar(X, Base.return_types(f, (eltype(X),)))
+        v = f(X.values[i])
+        dest = similar(X, typeof(v))
+        dest[i] = v
+        map_to!(f, dest, X, i+1, len)
+    end
+
+    function _liftedmap(f, X1::NullableArray, X2::NullableArray)
+        len = prod(promote_shape(X1, X2))
+        len > 0 || return NullableArray{Base.return_types(f, (eltype(X1), eltype(X2))), 0}()
+        i = 1
+        while X1.isnull[i] | X2.isnull[i]
+            i += 1
+        end
+        i <= len || return similar(X1, Base.return_types(f, (eltype(X1), eltype(X2))))
+        v = f(X1.values[i], X2.values[i])
+        dest = similar(X1, typeof(v))
+        dest[i] = v
+        map_to!(f, dest, X1, X2, i+1, len)
+    end
+
+    @generated function _liftedmap{N}(f, Xs::Vararg{NullableArray, N})
+        return quote
+            shp = size(zip(Xs...))
+            len = prod(shp)
+            i = 1
+            while @nullcheck Xs $N
+                i += 1
+            end
+            i <= len || return similar(X1, Base.return_types(f, tuple([ eltype(X) for X in Xs ])))
+            v = @fcall Xs $N
+            dest = similar(Xs[1], typeof(v))
+            dest[i] = v
+            map_to!(f, dest, Xs, i+1, len)
+        end
+    end
+
+    function map_to!{T}(f, dest::NullableArray{T}, X, offs, len)
+        # map to dest array, checking the type of each result. if a result does not
+        # match, widen the result type and re-dispatch.
+        i = offs
+        while i <= len
+            @inbounds if X.isnull[i]
+                i += 1; continue
+            end
+            @inbounds el = f(X.values[i])
+            S = typeof(el)
+            if S === T || S <: T
+                @inbounds dest[i] = el::T
+                i += 1
+            else
                 R = typejoin(T, S)
                 new = similar(dest, R)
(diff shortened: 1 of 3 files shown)


nalimilan:
I wouldn't worry too much about 0.4 compatibility, keeping the existing code in a separate file is fine.

Regarding lifting or not, that's a tough call. In practice, I guess most people will use the lifting behavior, but doing this by default would mean we break the `AbstractArray` API by not applying the function on elements of the type of `eltype(x)`. Performance-wise, the benchmarks show a lot of allocations which I wouldn't expect: are you sure there isn't a problem somewhere? (I don't master that code at all.)

Finally, I wonder about the difference between these two benchmarks:

``` julia

julia> @benchmark map(f, As...)
BenchmarkTools.Trial: 
  samples:          158
  evals/sample:     1
  time tolerance:   5.00%
  memory tolerance: 1.00%
  memory estimate:  7.63 mb
  allocs estimate:  10
  minimum time:     27.83 ms (0.00% GC)
  median time:      30.86 ms (0.00% GC)
  mean time:        31.65 ms (2.66% GC)
  maximum time:     40.85 ms (6.13% GC)

julia> @benchmark map(f, Xs...; lift=true)
BenchmarkTools.Trial: 
  samples:          354
  evals/sample:     1
  time tolerance:   5.00%
  memory tolerance: 1.00%
  memory estimate:  8.58 mb
  allocs estimate:  32
  minimum time:     12.27 ms (0.00% GC)
  median time:      13.62 ms (0.00% GC)
  mean time:        14.18 ms (6.01% GC)
  maximum time:     23.19 ms (10.96% GC)
```

How it is that the `NullableArray` version is twice faster? :-)

davidagold:
@nalimilan 

w/r/t the allocations on non-lifted `map`: I thought these are due to all the `Nullable` objects being created by indexing. Or should they not show up? There may be a problem.

w/r/t beating `map(f, As...)`: Haha, well... there are two primary differences (in my eyes) between the implementation of `map(f, As...)` here and in Base. The first is that the Base version `collect`s over `Generator(f, As...)`, and the latter `zip`s all of the `As` together in a `Zip` object. I don't know if this introduces overhead, but it might be less efficient than working with the `As` directly, as we do here. The second difference has to do with indexing. The Base implementation uses `ith_all(i, As)...` to access the ith element of each argument `Array`, which involves a fair amount of splatting. What we do here is use a macro `@fcall` to splice `f(Xs[1][i], Xs[2][i], Xs[3][i], ...)` into the `_liftedmap_to!` loop body. This does require that `_liftedmap_to!` be a generated function.

Whatever the case, if you write `map` for regular `Array`s in the style I've written them for `NullableArray`s, you do see a speed up over the current Base implementation: https://gist.github.com/davidagold/d7088aae22f23d383e5bf1f26aa1a045

``` julia
using BenchmarkTools
n = 1_000_000
As = [ rand(n) for i in 1:5 ]
f(xs...) = prod(xs)

julia> @benchmark mymap(f, As...)
BenchmarkTools.Trial: 
  samples:          786
  evals/sample:     1
  time tolerance:   5.00%
  memory tolerance: 1.00%
  memory estimate:  7.63 mb
  allocs estimate:  8
  minimum time:     4.59 ms (0.00% GC)
  median time:      5.35 ms (0.00% GC)
  mean time:        6.36 ms (14.65% GC)
  maximum time:     34.15 ms (0.00% GC)

julia> mymap(f, As...) == map(f, As...)
true
```

This suggests that the current Base n-arg `map` implementation leaves some performance on the table --- around 5x in this case.

nalimilan:
> w/r/t the allocations on non-lifted map: I thought these are due to all the Nullable objects being created by indexing. Or should they not show up? There may be a problem.

Not sure. But since it's an immutable, it should be possible to avoid the allocation at least in some scenarios. For example, by inlining the function which creates the `Nullable` (the rest of the work could be wrapped in a `@noinline` function). You'd need to ask more expert people about that.

As regards being faster than Base's `map`: could you open an issue to discuss this, or ask for feedback on Slack?

davidagold:
I'm going to merge this, since it passes on 0.4 and the `map` tests pass on `nightly` -- it's just the other stuff that's broken due to the introduction of the `Base.OneTo` range object that's causing tests to fail. The design choices above and their implementation don't need to be the final word, but I think they're a step in the right direction. And I feel it's important to fix #130.

ararslan:
> or ask for feedback on Slack

There's a Slack group?

johnmyleswhite:
A few of us chat on a private Slack channel.

ararslan:
Ah, okay. Org owners only, I presume?

Issue #125, opened 2016-06-27
Missing constructors


davidagold:
On latest master:

``` julia
julia> Array{Int}()
0-dimensional Array{Int64,0}:
4642688496

julia> NullableArray{Int}()
ERROR: MethodError: no method matching NullableArrays.NullableArray{Int64,N}()
 in eval(::Module, ::Any) at ./boot.jl:234
 in macro expansion at ./REPL.jl:92 [inlined]
 in (::Base.REPL.##1#2{Base.REPL.REPLBackend})() at ./event.jl:46

julia> NullableArray{Int}(5)
ERROR: MethodError: Cannot `convert` an object of type Int64 to an object of type NullableArrays.NullableArray{Int64,N}
This may have arisen from a call to the constructor NullableArrays.NullableArray{Int64,N}(...),
since type constructors fall back to convert methods.
 in NullableArrays.NullableArray{Int64,N}(::Int64) at ./sysimg.jl:63
 in eval(::Module, ::Any) at ./boot.jl:234
 in macro expansion at ./REPL.jl:92 [inlined]
 in (::Base.REPL.##1#2{Base.REPL.REPLBackend})() at ./event.jl:46
```

quinnj:
Just ran into this myself. +1

Issue #124, opened 2016-06-27
Error showing empty NullableArrays


davidagold:
On version `0.5.0-dev+4925`, with current NullableArrays master:

``` julia
julia> NullableArray{Int,0}()
0-dimensional NullableArrays.NullableArray{Int64,0}:
Error showing value of type NullableArrays.NullableArray{Int64,0}:
ERROR: StackOverflowError:
 in indexed_next(::Tuple{Symbol,Bool}, ::Int64, ::Int64) at ./tuple.jl:33
 in #IOContext#154(::Array{Any,1}, ::Type{T}, ::IOContext{Base.Terminals.TTYTerminal}) at ./show.jl:36
 in (::Core.#kw#Type)(::Array{Any,1}, ::Type{IOContext}, ::IOContext{Base.Terminals.TTYTerminal}) at ./null:0
 in showarray(::IOContext{Base.Terminals.TTYTerminal}, ::NullableArrays.NullableArray{Int64,0}) at ./show.jl:1544
 in show(::IOContext{Base.Terminals.TTYTerminal}, ::NullableArrays.NullableArray{Int64,0}) at ./show.jl:1537
 in showarray(::IOContext{Base.Terminals.TTYTerminal}, ::NullableArrays.NullableArray{Int64,0}) at ./show.jl:1557
 in show(::IOContext{Base.Terminals.TTYTerminal}, ::NullableArrays.NullableArray{Int64,0}) at ./show.jl:1537
...
```

PR #133, opened 2016-07-23
Address issues #124, #125

davidagold:
Includes tests.

Worth noting that the bug in #124 involved how we treated `getindex(X::NullableArray)`. A reminder that we could stand to review our general indexing strategy and its coordination with that in Base for `AbstractArray`s.


davidagold:
Failure on nightly is due to unrelated issues with `broadcast`, which will be addressed in a separate PR.

Issue #132, opened 2016-07-22
empty! not defined for NullableArrays


quinnj:

davidagold:
added in #135

PR #135, opened 2016-07-25
Add empty! for NullableVector

davidagold:
Address #132


ararslan:
What about for arrays that aren't one-dimensional (i.e. aren't vectors)?

davidagold:
I assume it's because there's no use in obtaining an empty higher-dim array because you can't push to it.

ararslan:
Ah yeah, good point. Looks good to me then!

PR #138, opened 2016-08-09
Add appropriate no-op convert method

davidagold:
`convert{T,N}(::Type{NullableArray}, X::NullableArray{T,N})` should be a no-op.


davidagold:
Tests pass on 0.4, 0.5, and the failure on nightly appears to be unrelated. That's good enough for me for now -- we need to tag a new version for 0.5.

PR #139, opened 2016-08-27
Add reshape(::NullableArray, shp) method

nalimilan:
Without this, the fallback ReshapedArray is used, which is unnecessarily
complex, and creates issues e.g. when trying to convert the result to an
Array{T}.

---

The interaction between `NullableArrays`, `reshape` and `convert` currently doesn't work very well. While NullableArrays.jl provides methods allowing conversion to `Array`, these do not apply to reshaped `NullableArrays`. In particular, this means one cannot reshape a `NullableVector` to a matrix in order to convert it to a `Matrix` (which is needed e.g. to build model matrices):

``` julia
julia> using NullableArrays

julia> convert(Matrix{Int}, reshape(NullableArray([1,2]), 2, 1))
ERROR: MethodError: Cannot `convert` an object of type Nullable{Int64} to an object of type Int64
This may have arisen from a call to the constructor Int64(...),
since type constructors fall back to convert methods.
 in copy!(::Base.LinearFast, ::Array{Int64,2}, ::Base.LinearFast, ::Base.ReshapedArray{Nullable{Int64},2,NullableArrays.NullableArray{Int64,1},Tuple{}}) at ./abstractarray.jl:551
 in convert(::Type{Array{Int64,2}}, ::Base.ReshapedArray{Nullable{Int64},2,NullableArrays.NullableArray{Int64,1},Tuple{}}) at ./array.jl:233

```

This PR fixes the above error.

### DIFF
--- a/src/primitives.jl
+++ b/src/primitives.jl
@@ -108,6 +108,10 @@ function Base.resize!{T}(X::NullableArray{T,1}, n::Int) # -> NullableArray{T, 1}
     return X
 end
 
+function Base.reshape(X::NullableArray, dims::Dims) # -> NullableArray
+    NullableArray(reshape(X.values, dims), reshape(X.isnull, dims))
+end
+
 @doc """
 `ndims(X::NullableArray)`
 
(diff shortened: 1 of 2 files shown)


nalimilan:
Failure on nightlies seem unrelated.

nalimilan:
Bump! This should be fairly uncontroversial.

davidagold:
Sorry for the delay.

PR #140, opened 2016-08-28
Widen dropnull() and allnull() signatures

nalimilan:
Support for any AbstractArray is needed to write generic code supporting
e.g. Array{Nullable} and NullableCategoricalArray. Make the semantics
of dropnull() more explicit in the docs.

Simplify the implementation of anynull(). In dropnull(X::NullableVector),
remove the redundant copy(): indexing already makes a copy.

(Travis failure on master is unrelated.)

### DIFF
--- a/src/primitives.jl
+++ b/src/primitives.jl
@@ -148,51 +148,50 @@ function Base.find(X::NullableArray{Bool}) # -> Array{Int}
     return res
 end
 
-@doc """
-`dropnull(X::NullableVector)`
 
-Return a `Vector` containing only the non-null entries of `X`.
+_isnull(x::Any) = false
+_isnull(x::Nullable) = isnull(x)
+
+@doc """
+`dropnull(X::AbstractVector)`
+
+Return a vector containing only the non-null entries of `X`,
+unwrapping `Nullable` entries. A copy is always returned, even when
+`X` does not contain any null values.
 """ ->
-dropnull(X::NullableVector) = copy(X.values[!X.isnull]) # -> Vector{T}
+function dropnull(X::AbstractVector)                 # -> AbstractVector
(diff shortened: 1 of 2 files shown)


nalimilan:
@davidagold What do you think about this one?

davidagold:
Thank you for pinging me. I think that these expanded definitions should live elsewhere, but I also think that getting them into Base (or wherever else we decide they ought to go) may take a while. So until then, I think they can live here. Does that sound right to you?

nalimilan:
I'd like to try moving into Base the extended definition `isnull(::Any) = false`. As regards `allnull`, `anynull` and `dropnull`, they are certainly more general than `NullableArrays`, but OTOH they are not very likely to be used outside of it since `Array{Nullable}` is quite inefficient. So I wouldn't say they are the most urgent functions to move to another place: first, we should move functions that apply to `Nullable`.

davidagold:
+2 to moving `isnull(::Any) = false` to Base. I do think that there's a case for `dropnull` to live in `base/nullable.jl`, but I'm fine with housing these methods here for now.

EDIT: The reasoning being: this isn't a semantics-related decision, and we can always move these definitions around later. No point in holding up progress elsewhere.

nalimilan:
Thanks! Could you (or I) tag a new release so that the DataFrames port passes tests?

davidagold:
Please feel free to tag whenever =)

quinnj:
+3 to having `isnull(::Any) = false` to Base. @johnmyleswhite, any hesitations with doing that?

johnmyleswhite:
No, I think that's good as long as we make sure it's getting optimized out for non-Nullable types.

nalimilan:
OK, I'll make a PR then.

PR #141, opened 2016-08-28
Add isless(::Nullable, ::Nullable) -> Bool

nalimilan:
This method complements isequal(), which also returns a Bool, as it
must implement a total order. NULLs are treated like NaNs: they are
never isless() than any other value, including other NULLs. They are
therefore always sorted last.

(Travis failure on master is unrelated.)

### DIFF
--- a/src/operators.jl
+++ b/src/operators.jl
@@ -1,7 +1,7 @@
 ## Lifted operators
 
 importall Base.Operators
-import Base: promote_op, abs, abs2, sqrt, cbrt, scalarmin, scalarmax
+import Base: promote_op, abs, abs2, sqrt, cbrt, scalarmin, scalarmax, isless
 using Compat: @functorize
 
 # Methods adapted from Base Julia 0.5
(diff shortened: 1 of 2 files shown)


davidagold:
This seems reasonable. But do you have a sense of how many more of these "special cases" (i.e., functions on `Nullable`s that do not return `Nullable`) are needed?

nalimilan:
Very few AFAICT. Looks like we all agree that all operators should return a `Nullable`, except for `isequal` and `isless`, and comparison operators (`==`, `<`, `>`) for which there's no consensus yet. So after this PR (which I'll open against Julia Base too) it should be complete.

davidagold:
Yeah, I'd prefer that this live in `Base`. Could you please x-ref that PR when you open it?

nalimilan:
I'd still like this to be included in NullableArrays too so that it works on 0.4 and 0.5. Requiring 0.6 is quite cumbersome at this point.

davidagold:
Fair, but can we organize the files so that this method can be grouped amongst those that should really live somewhere else?

nalimilan:
I think that's the case. Aren't `Nullable` operators supposed to go into Base or into a separate package at some point? Also, this definition is very dependent on the infrastructure used for other operators.

davidagold:
Sounds good to me for now.

nalimilan:
Filed PR against Base: https://github.com/JuliaLang/julia/pull/18304

PR #146, opened 2016-09-17
Avoid a DomainError if v2 ever happens to be typemin of a signed inte…

tkelman:
…ger type

e.g. `abs(Int8(-128)) == -128`

ref https://github.com/JuliaLang/julia/pull/18478#issuecomment-246877142 - I'm not sure why changing `VERSION` in Julia would result in different random numbers here despite the `srand` in this file. Any ideas why that would happen?

### DIFF
--- a/test/operators.jl
+++ b/test/operators.jl
@@ -78,7 +78,7 @@ module TestOperators
         v1 = one(T)
         v2 = T <: Union{BigInt, BigFloat} ? T(rand(Int128)) : rand(T)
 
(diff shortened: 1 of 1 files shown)


tkelman:
Nightly looks like it was failing when 0.0.8 was tagged a few weeks ago. Reason might be slightly different, but the red status isn't new.

davidagold:
I've no idea what's behind the RNG behavior. @nalimilan ? Thanks for the patch, Tony.

nalimilan:
No idea either, but looks good to me.

tkelman:
Thanks for the review and merge! Mind tagging?

PR #150, opened 2016-09-24
Prepare for change from isnull to hasvalue field in Base

nalimilan:
So that we're ready for https://github.com/JuliaLang/julia/pull/18510. Also include a small fix for `isless`.

Tests fail on 0.6 because of issues that existed before.

### DIFF
--- a/README.md
+++ b/README.md
@@ -145,7 +145,7 @@ Let `v, w` be two `Nullable{Float64}` objects. If both `v, w` are non-null, the 
 Arguably, the best way to lift existing methods over `Nullable` arguments is to use multiple dispatch. That is, one can very easily extend `f` to handle `Nullable{Float64}` arguments by simply defining an appropriate method:
 ```julia
 function f(x::Nullable{Float64}, y::Nullable{Float64})
-    if x.isnull | y.isnull
+    if isnull(x) | isnull(y)
         return Nullable{Float64}()
     else
         return Nullable(f(x.value, y.value))
--- a/perf/reduce.jl
+++ b/perf/reduce.jl
@@ -12,7 +12,8 @@ D = DataArray(A)
 E = DataArray(A, B)
 
 f(x) = 5 * x
-f{T<:Number}(x::Nullable{T}) = Nullable(5 * x.value, x.isnull)
+f{T<:Number}(x::Nullable{T}) =
+    ifelse(isnull(x), Nullable{typeof(5 * x.value)}(), Nullable(5 * x.value))
 
 #-----------------------------------------------------------------------------#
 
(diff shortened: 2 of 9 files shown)


README.md
```diff
@@ -145,7 +145,7 @@ Let `v, w` be two `Nullable{Float64}` objects. If both `v, w` are non-null, the
 Arguably, the best way to lift existing methods over `Nullable` arguments is to use multiple dispatch. That is, one can very easily extend `f` to handle `Nullable{Float64}` arguments by simply defining an appropriate method:
```
davidagold, line 145: I suppose I did write this. I should change this README, since now I believe precisely the opposite. =p

perf/reduce.jl
```diff
@@ -12,7 +12,7 @@ D = DataArray(A)
 E = DataArray(A, B)
 
 f(x) = 5 * x
-f{T<:Number}(x::Nullable{T}) = Nullable(5 * x.value, x.isnull)
+f{T<:Number}(x::Nullable{T}) = ifelse(isnull(x), Nullable{typeof(5*x.value)}, Nullable(5 * x.value))
```
davidagold: Is the `typeof` bit necessary?

src/operators.jl
```diff
@@ -4,6 +4,12 @@ importall Base.Operators
 import Base: promote_op, abs, abs2, sqrt, cbrt, scalarmin, scalarmax, isless
 using Compat: @functorize
 
+if isdefined(Base, :fieldname) && Base.fieldname(Nullable, 1) == :hasvalue # Julia 0.6
+    Nullable(R, x, hasvalue::Bool) = Nullable{R}(x, hasvalue)
```
davidagold: Do we need these? I'm loathe to define new methods for `Nullable`, including constructors, in this package.

davidagold:
Pesky `broadcast`.

Should we change the `NullableArray` internals to reflect this change? I'm not sure I like the disparity.

perf/reduce.jl
nalimilan: Yes, at least if we want to ensure type stability in all cases. But I just realized this is missing `()`, and the fact that tests haven't failed means that it's not really tested on nulls. Will fix.

src/operators.jl
nalimilan: I can call this `_Nullable` if you prefer to ensure it's not exported.

nalimilan:
> Should we change the NullableArray internals to reflect this change? I'm not sure I like the disparity.

Yeah, but we can do that at any point independently of changes in Base.

src/operators.jl
davidagold: Is there any reason we can't just use `Nullable{R}(x, hasvalue)`?

nalimilan: Because on 0.4 and 0.5 the second argument is `isnull`, but on 0.6 it will (probably) be `hasvalue`.

PR #151, opened 2016-09-29
AbstractNullableArray/Vector/Matrix typealiases

alyst:
`AbstractArray` counterparts. Should be handy for declaring generic methods.


alyst:
don't know why nightly build failed

PR #152, opened 2016-10-03
implemented NullableArray{T,N}(dims...)

gustafsson:
This is needed for instance when I have a NullableArray `A` of some type and want to create another one of the same type with only nulls in it like so: `typeof(A)(size(A)...)`

### DIFF
--- a/src/constructors.jl
+++ b/src/constructors.jl
@@ -34,6 +34,7 @@ end
 
 @compat (::Type{NullableArray{T}}){T}(dims::Dims) = NullableArray(T, dims)
 @compat (::Type{NullableArray{T}}){T}(dims::Int...) = NullableArray(T, dims)
+@compat (::Type{NullableArray{T,N}}){T,N}(dims::Vararg{Int,N}) = NullableArray(T, dims)
 
 # The following method constructs a NullableArray from an Array{Any} argument
 # 'A' that contains some placeholder of type 'T' for null values.
--- a/src/constructors.jl
+++ b/src/constructors.jl
@@ -34,14 +34,13 @@ end
 
 @compat (::Type{NullableArray{T}}){T}(dims::Dims) = NullableArray(T, dims)
 @compat (::Type{NullableArray{T}}){T}(dims::Int...) = NullableArray(T, dims)
-if VERSION >= v"0.5.0-dev"
-    (::Type{NullableArray{T,N}}){T,N}(dims::Vararg{Int,N}) = NullableArray(T, dims)
+if VERSION >= v"0.5.0-"
+    @compat (::Type{NullableArray{T,N}}){T,N}(dims::Vararg{Int,N}) = NullableArray(T, dims)
 else
-    @compat (::Type{NullableArray{T,1}}){T,N}(x1) = NullableArray(T, x1)
-    @compat (::Type{NullableArray{T,2}}){T,N}(x1, x2) = NullableArray(T, x1, x2)
-    @compat (::Type{NullableArray{T,3}}){T,N}(x1, x2, x3) = NullableArray(T, x1, x2, x3)
-    @compat (::Type{NullableArray{T,3}}){T,N}(x1, x2, x3, x4) = NullableArray(T, x1, x2, x3, x4)
-    @compat (::Type{NullableArray{T,3}}){T,N}(x1, x2, x3, x4, x5) = NullableArray(T, x1, x2, x3, x4, x5)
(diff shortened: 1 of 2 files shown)


src/constructors.jl
```diff
@@ -34,6 +34,15 @@ end
 
 @compat (::Type{NullableArray{T}}){T}(dims::Dims) = NullableArray(T, dims)
 @compat (::Type{NullableArray{T}}){T}(dims::Int...) = NullableArray(T, dims)
+if VERSION >= v"0.5.0-dev"
+    (::Type{NullableArray{T,N}}){T,N}(dims::Vararg{Int,N}) = NullableArray(T, dims)
+else
+    @compat (::Type{NullableArray{T,1}}){T,N}(x1) = NullableArray(T, x1)
+    @compat (::Type{NullableArray{T,2}}){T,N}(x1, x2) = NullableArray(T, x1, x2)
```
nalimilan: Why define these on 0.4 but not on 0.5? Performance?

davidagold:
Is there any reason you can't use `similar`?

``` julia
julia> A = collect(1:4)
4-element Array{Int64,1}:
 1
 2
 3
 4

julia> X = NullableArray(A)
4-element NullableArrays.NullableArray{Int64,1}:
 1
 2
 3
 4

julia> Y = similar(X)
4-element NullableArrays.NullableArray{Int64,1}:
 #NULL
 #NULL
 #NULL
 #NULL
```

src/constructors.jl
gustafsson: On 0.5 a single function takes care of all of them.

gustafsson:
@davidagold `similar` is a good idea. I could do something like `similar(T(),3,2,1)` when I only have the type of a NullableArray but not an instance. Is it possible to do without creating an empty instance of T (perhaps it gets optimised away anyways?).

gustafsson:
I don't need this anymore.

nalimilan:
I think it would still be good to have.

gustafsson:
@nalimilan ok, I fixed the errors on v"0.4"

src/constructors.jl
```diff
@@ -34,6 +34,14 @@ end
 
 @compat (::Type{NullableArray{T}}){T}(dims::Dims) = NullableArray(T, dims)
 @compat (::Type{NullableArray{T}}){T}(dims::Int...) = NullableArray(T, dims)
```
nalimilan, line 36: Is this definition still useful?

gustafsson: It's needed in order to do things like `NullableArray{Int}(3,2)`.

But a call to `NullableArray{T}(dims::Int...)` such as `A(dim1,dim2)` can always be replaced by `A((dim1,dim2))` to call the other constructor with `::Dims` instead. And `A(dims...)` can be replaced with `A((dims...))`. I think it is convenient keep this though.

nalimilan: Ah, right, I thought `N` could be inferred automatically in your new constructor, but that doesn't work for the call overloading syntax.

nalimilan (approved):

nalimilan:
@davidagold Good to merge?

src/constructors.jl
```diff
@@ -34,6 +34,14 @@ end
...
+if VERSION >= v"0.5.0-"
+    @compat (::Type{NullableArray{T,N}}){T,N}(dims::Vararg{Int,N}) = NullableArray(T, dims)
+else
+  function Base.convert{T,N}(::Type{NullableArray{T,N}}, dims::Int...)
+      length(dims) == N || throw(ArgumentError("Wrong number of arguments. Expected $N, got $(length(dims))."))
+      NullableArray(T, dims)
+  end
```
ararslan: Please use 4 space indents for consistency with the rest of the repo.

ararslan (changes requested):

davidagold:
What's the failure?

ararslan (approved):

ararslan:
Changed my review since my comments were addressed, but I think the cause of the test failure should be determined before this is merged.

gustafsson:
I don't see how the error is related to this PR. v0.4 and v0.5 pass. but not nightly. does the tests currently pass on master?

nalimilan:
The error is unrelated, I'm able to reproduce it on master too. Trying to figure out what's going on...

EDIT: that's a bug in Base, see https://github.com/JuliaLang/julia/issues/19270.

nalimilan:
All is green now, merging.

nalimilan:
Damn, I screwed up and forgot to squash. Squashed and force-pushed manually.

PR #155, opened 2016-10-04
[RFC/WIP] Few additional methods

alyst:
Some methods that I introduced, while rebasing JuliaStats/DataFrames.jl#850
- `isnull(A::NullableArray)` to get the boolean array of `A` nulls
- `unsafe_isnull(A::NullableArray, i::Int...)` to check for nulls without checking the validity of the index
- `AbstractNullableArray{T, N}` typealias of `AbstractArray{Nullable{T}, N}` (however `AbstractNullableArray` (without params) is not considered by Julia more specific than `AbstractArray`, I don't know if it's a known feature/bug)
-  `isless{T,S}(a::Nullable{T}, b::S)` to compare nullables vs nonnullables (nonnullable is always less)

### DIFF
--- a/src/indexing.jl
+++ b/src/indexing.jl
@@ -77,18 +77,18 @@ to store `v` at `I`.
 end
 
 # return the value of non-null X element wrapped in Nullable
-function unsafe_getindex_notnull(X::NullableArray, I::Int...)
+@inline function unsafe_getindex_notnull(X::NullableArray, I::Int...)
     return Nullable(getindex(X.values, I...))
 end
-function unsafe_getindex_notnull{T}(X::AbstractArray{Nullable{T}}, I::Int...)
+@inline function unsafe_getindex_notnull{T}(X::AbstractArray{Nullable{T}}, I::Int...)
     return getindex(X, I...)
 end
 
 # return the value of non-null X element
-function unsafe_getvalue_notnull(X::NullableArray, I::Int...)
+@inline function unsafe_getvalue_notnull(X::NullableArray, I::Int...)
     return getindex(X.values, I...)
 end
-function unsafe_getvalue_notnull{T}(X::AbstractArray{Nullable{T}}, I::Int...)
+@inline function unsafe_getvalue_notnull{T}(X::AbstractArray{Nullable{T}}, I::Int...)
     return get(getindex(X, I...))
 end
 
@@ -142,6 +142,6 @@ end
 
 This is a convenience method to set the entry of `X` at index `I` to be null
 """
-function nullify!(X::NullableArray, I...)
+@inline function nullify!(X::NullableArray, I...)
     setindex!(X, Nullable{eltype(X)}(), I...)
 end
--- a/src/indexing.jl
+++ b/src/indexing.jl
@@ -76,10 +76,12 @@ to store `v` at `I`.
     return v
 end
 
+# return the value of non-null X element wrapped in Nullable
 function unsafe_getindex_notnull(X::NullableArray, I::Int...)
     return Nullable(getindex(X.values, I...))
 end
 
+# return the value of non-null X element
 function unsafe_getvalue_notnull(X::NullableArray, I::Int...)
     return getindex(X.values, I...)
 end
--- a/src/indexing.jl
+++ b/src/indexing.jl
@@ -80,10 +80,16 @@ end
 function unsafe_getindex_notnull(X::NullableArray, I::Int...)
     return Nullable(getindex(X.values, I...))
 end
+function unsafe_getindex_notnull{T}(X::AbstractArray{Nullable{T}}, I::Int...)
+    return getindex(X, I...)
+end
 
 # return the value of non-null X element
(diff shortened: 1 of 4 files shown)


nalimilan:
> isnull(A::NullableArray) to get the boolean array of A nulls

This conflicts with the new definition of `isnull(::Any)` in Julia 0.6. I'd prefer using `isnull.(A)`, which should work if we fix `broadcast`.

> unsafe_isnull(A::NullableArray, i::Int...) to check for nulls without checking the validity of the index

Can't we use the `@inbounds` machinery combined with `isnull(A, i)` for that?

> isless{T,S}(a::Nullable{T}, b::S) to compare nullables vs nonnullables (nonnullable is always less)

Why would you want this sort order? I'd rather use lifting semantics after wrapping `b` inside a `Nullable`, but we shouldn't do that without a clear plan of what functions need this. EDIT: reading the code, I see that only null values are sorted last. So that's just an extension of `isless` to mixed cases. I would be on board with that, but we should also do the same for `isequal`, as well as for all operators. This deserves a separate issue (see https://github.com/JuliaStats/NullableArrays.jl/pull/85).

alyst:
> > isnull(A::NullableArray) to get the boolean array of A nulls
> 
> This conflicts with the new definition of isnull(::Any) in Julia 0.6. I'd prefer using isnull.(A), which should work if we fix broadcast.

Would `isnull.()` be able to just return `A.isnull`?

> > unsafe_isnull(A::NullableArray, i::Int...) to check for nulls without checking the validity of the index
> 
> Can't we use the @inbounds machinery combined with isnull(A, i) for that?

I must admit my understanding of what `@inbounds` can do is very limited. Would it see across the function boundary? The other problem is that `@inbounds Expr` IIRC returns nothing, which is sometimes inconvenient.

> > isless{T,S}(a::Nullable{T}, b::S) to compare nullables vs nonnullables (nonnullable is always less)
> 
> ... I would be on board with that, but we should also do the same for isequal, as well as for all operators. This deserves a separate issue (see #85).

Actually, this one I don't really need, so I will remove it from the PR. Anyway, if it has to be defined, it should go to `Base`.

nalimilan:
> Would isnull.() be able to just return A.isnull?

It would, by overloading `broadcast(::typeof(isnull), ::NullableArray)`. Though we should check whether it's OK to return the internal field without a copy. It could certainly be confusing.

> I must admit my understanding of what @inbounds can do is very limited. Would it see across the function boundary? The other problem is that @inbounds Expr IIRC returns nothing, which is sometimes inconvenient.

You just need to add `@boundscheck` at the beginning of the line which does the bounds checking, and mark the `getindex` function as `@inline`. See http://docs.julialang.org/en/release-0.5/devdocs/boundscheck/. Do you have an example where the fact that `@inbounds` doesn't return anything would really be a problem?

davidagold:
> Though we should check whether it's OK to return the internal field without a copy. It could certainly be confusing.

Agreed. I think I'd prefer a copy. This might be ameliorated by switching over to a `:hasvalue` field for `NullableArray`s and then have only a generic method for `isnull` but not one for `hasvalue` so that `hasvalue(X::NullableArray)` could select the `:hasvalue` field. 

EDIT: Or we could just require that anybody who wants to modify the internals of a `NullableArray` do their own dirty work and explicitly use `getfield`. I think I like that the best.

alyst:
Is there any possibility to return a readonly view of the `A.isnull`?

davidagold:
I don't think so. That seems like a weird special case. What's wrong with `Base.getfield`?

alyst:
@davidagold I thought the reason to return a copy of `A.isnull` is to avoid accidental modifications of the field. OTOH in most of the cases `isnull(A)` would be used for reading only, and copying would be unnecessary performance penalty. It would be nice if `isnull(A)` returns something that implements `AbstractArray{Bool}` interface without `setindex!()` and doesn't allocate. Then the same could be done for `NullableCategoricalArray`, which doesn't have the `:isnull` field.

nalimilan:
Anyway as David said we need to change `isnull` to `hasvalue` field for consistency with e.g. Arrow, Pandas 2 or PostgreSQL. I guess we could return a generator, like `isnull(x::NullableArray) = (!x.hasvalue[i] for i in eachindex(x.hasvalue))`. That would also work for CategoricalArrays. But it could be surprising for users...

cjprybol:
bump re: `hasvalue`

> Anyway as David said we need to change isnull to hasvalue field for consistency with e.g. Arrow, Pandas 2 or PostgreSQL. I guess we could return a generator, like isnull(x::NullableArray) = (!x.hasvalue[i] for i in eachindex(x.hasvalue)). That would also work for CategoricalArrays. But it could be surprising for users...

nalimilan:
I still think moving to `hasvalue` is a good idea in general, but given the expected changes in the representation of `Nullable` in Base (as a `Union`), and given that changing that field won't really improve the user experience, I'd rather wait until the future of NullableArrays is clearer. We don't really need more disruptive changes at the moment.

Issue #159, opened 2016-10-06
Update docstrings to new conventions 


nalimilan:
We should improve our docstrings by adapting them to new conventions [from the Julia manual](http://docs.julialang.org/en/latest/manual/documentation/). In particular (not all of these steps need to be done at the same time, though):
 - [x] stop using `@doc` and `->` (except for special cases)
 - [x] give the signature with a four-space indent (instead of backquotes)
 - [ ] use imperative form ("return") instead of third person ("returns")
 - [ ] remove docstrings for generic functions which do not really add `NullableArray`-specific information (like `unshift!`, `shift!`, `pop!`, `splice!`, etc.)
 - [ ] add examples as doctests where useful

nalimilan:
Points 1 and 2 have been addressed by https://github.com/JuliaStats/NullableArrays.jl/pull/162 (thanks, @RatanRSur!).

ararslan:
I hope you don't mind but I converted your list to a checklist for easier tracking of what's been done.

PR #162, opened 2016-10-13
update docstrings to new format

RatanRSur:
Shamelessly took the low hanging fruit :)

### DIFF
--- a/src/broadcast.jl
+++ b/src/broadcast.jl
@@ -68,8 +68,9 @@ if VERSION < v"0.5.0-dev+5434"
     end
 
     @eval let cache = Dict{Any, Dict{Bool, Dict{Int, Dict{Int, Any}}}}()
-        @doc """
-        `broadcast!(f, B::NullableArray, As::NullableArray...; lift::Bool=false)`
+        """
+            broadcast!(f, B::NullableArray, As::NullableArray...; lift::Bool=false)
+
         This method implements the same behavior as that of `broadcast!` when called on
         regular `Array` arguments. It also includes the `lift` keyword argument, which
         when set to true will lift `f` over the entries of the `As`.
@@ -79,7 +80,7 @@ if VERSION < v"0.5.0-dev+5434"
         `NullableArray`s. Thus, calling `broadcast!` on a arguments consisting
         of both `Array`s and `NullableArray`s will fall back to the implementation
         of `broadcast!` in `base/broadcast.jl`.
-        """ ->
+        """
         function Base.broadcast!(f, B::NullableArray, As::NullableArray...; lift::Bool=false)
             nd = ndims(B)
             narrays = length(As)
@@ -148,8 +149,8 @@ else
         end
     end
 
-    @doc """
-    `broadcast!(f, B::NullableArray, As::NullableArray...; lift::Bool=false)`
+    """
+        broadcast!(f, B::NullableArray, As::NullableArray...; lift::Bool=false)
 
     This method implements the same behavior as that of `broadcast!` when called
     on regular `Array` arguments. It also includes the `lift` keyword argument,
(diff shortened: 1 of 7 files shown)


ararslan:
Thanks for the contribution!

As an aside, it's best not to use the `@` symbol at all in commit messages, since it actually ends up sending a notification to the user with that name. In this case, [this person](https://github.com/doc) has gotten or will get a notification. I think the common practice is to say `at-doc` or `doc macro`, etc. in commit messages. If you'd like, you can `git commit --amend` and then force push to your branch.

So that GitHub will auto-close the issue once this is merged: Fixes #159

src/broadcast.jl
```diff
@@ -68,8 +68,8 @@ if VERSION < v"0.5.0-dev+5434"
     end
 
     @eval let cache = Dict{Any, Dict{Bool, Dict{Int, Dict{Int, Any}}}}()
-        @doc """
-        `broadcast!(f, B::NullableArray, As::NullableArray...; lift::Bool=false)`
+        """
+            `broadcast!(f, B::NullableArray, As::NullableArray...; lift::Bool=false)`
```
ararslan: When these are indented, the backticks should be removed.

ararslan (changes requested):

RatanRSur:
Oh, cool I did not know that about @'s. Thanks!

Let me know if there are more changes.

src/indexing.jl
```diff
@@ -129,8 +129,8 @@ function Base.to_index(X::NullableArray)
     Base.to_index(X.values)
 end
 
-@doc """
-`nullify!(X::NullableArray, I...)`
+"""
+   nullify!(X::NullableArray, I...)
 
 This is a convenience method to set the entry of `X` at index `I` to be null
 """ ->
```
ararslan: The arrows at the end of the doc strings should also go; they aren't part of the new syntax.

RatanRSur: Done, also I noticed I messed up and used 3 space indents so I fixed that as well

ararslan (approved):

ararslan:
@davidagold Is the nightly failure a known issue?

src/broadcast.jl
```diff
@@ -68,8 +68,8 @@ if VERSION < v"0.5.0-dev+5434"
     end
 
     @eval let cache = Dict{Any, Dict{Bool, Dict{Int, Dict{Int, Any}}}}()
-        @doc """
-        `broadcast!(f, B::NullableArray, As::NullableArray...; lift::Bool=false)`
+        """
+            broadcast!(f, B::NullableArray, As::NullableArray...; lift::Bool=false)
         This method implements the same behavior as that of `broadcast!` when called on
```
nalimilan, line 74: Missing line break here.

src/primitives.jl
```diff
@@ -112,30 +112,30 @@ function Base.reshape(X::NullableArray, dims::Dims) # -> NullableArray
...
+    endof(X::NullableArray)
 
 Returns the last entry of `X`.
-""" ->
+"""
 Base.endof(X::NullableArray) = endof(X.values) # -> Int
 
-@doc """
-
-""" ->
```
nalimilan, line 138: This empty comment is weird, probably time to remove it.

src/typedefs.jl
```diff
@@ -13,11 +13,11 @@
 #
 # TODO: Ensure that size(values) == size(isnull) using inner constructor.
 # TODO: Implement outer constructor required once we add an inner constructor.
-@doc """
-`NullableArray{T, N}` is an efficient alternative to `Array{Nullable{T}, N}`.
+"""
+    NullableArray{T, N} is an efficient alternative to `Array{Nullable{T}, N}`.
```
nalimilan: This line isn't a signature, it's just a sentence, so keep it as it was previously.

nalimilan (changes requested):
Thanks! Looks good apart from a few details.

nalimilan:
@ararslan The nightly failure is known, that's a problem with `broadcast`.

src/typedefs.jl
ararslan: Thanks for the thorough review, @nalimilan, and for catching some stuff I missed!

nalimilan (approved):

nalimilan:
Woops, completely forgot about this. It's no longer October, but let's pretend it is...

Issue #145, opened 2016-09-15
Constructor from Array{Nullable} gives eltype Nullable{Nullable}


nalimilan:
Looks like `Array{Nullable}` isn't handled like `Array{Nullable{T}}`:

``` julia
julia> nm = NullableArray(Nullable[Nullable() 0.0; 0.0 1.0])
2x2 NullableArrays.NullableArray{Nullable{T},2}:
 Nullable{Union{}}()  Nullable(0.0)
 Nullable(0.0)        Nullable(1.0)

```

davidagold:
Those should be consistent. And we should probably remark somewhere that the way to create a `NullableArray{Nullable{T}}` is by means of `NullableArray{Nullable{T}}(...)`.

PR #165, opened 2016-10-25
Fix constructor/convert for Array{Nullable} without element type

nalimilan:
Fixes https://github.com/JuliaStats/NullableArrays.jl/issues/145.


nalimilan:
Failures are the broadcast issue on nightlies only.

ararslan (approved):

Issue #163, opened 2016-10-21
`UndefVarError` when returning `Nullable` from `map` with `lift=true`


ExpandingMan:
``` julia
X = NullableArray(1:100)
map(x -> (x == 2 ? Nullable() : x), X, lift=true)
```

gives the following error

```
ERROR: UndefVarError: map_to! not defined
 in _liftedmap_to!(::##3#4, ::NullableArrays.NullableArray{Int64,1}, ::NullableArrays.NullableArray{Int64,1}, ::Int64, ::Int64) at /home/savastio/.julia/v0.5/NullableArrays/src/map.jl:198
 in _liftedmap(::##3#4, ::NullableArrays.NullableArray{Int64,1}) at /home/savastio/.julia/v0.5/NullableArrays/src/map.jl:140
 in #map#15 at /home/savastio/.julia/v0.5/NullableArrays/src/map.jl:113 [inlined]
 in (::Base.#kw##map)(::Array{Any,1}, ::Base.#map, ::Function, ::NullableArrays.NullableArray{Int64,1}) at ./<missing>:0
```

I suppose this may be something that simply hasn't been implemented yet, but if so it probably should have returned an error that said so.

Changing the returned `Nullable()` to be type specific does not help.

nalimilan:
I think the idea behind `lift=true` is that you're supposed to get non-null values and return non-null values. There's no reason not to support this use case, though.

ExpandingMan:
To give you an example of where it comes up: I decided that I wanted to convert all `NaN`s to `Nullable{Float64}()` for the sake of consistency.  To do this without lift one would need

``` julia
map(X) do x
    if !isnull(x) && isnan(get(x))
        return Nullable()
    end
    return x
end
```

Actually I think this is pretty typical of the design challenge facing this package: either most functions have to be overloaded to deal with `Nullable` (though there isn't always a perfectly consistent way of doing so) or there need to be iterators that make it easier for non-specialized functions to get applied.  It's probably not an unusual thing for non-null values to get converted to null, and in those cases you still don't really want to make special accommodations for existing nulls.

davidagold:
`lift=true` is just supposed to enable standard lifting semantics for whatever `f` is passed to `map` --- precisely the behavior that @ExpandingMan was expecting. I think that what is happening here is that, since there are no nulls in `X`, `map` is trying to fall back to an internal `map_to!` implementation from Base that, evidently, no longer exists.

[...]

Issue #144, opened 2016-08-29
Behavior of map() and broadcast()


nalimilan:
Currently, `map`/`broadcast` on `NullableArray` returns a `NullableArray`. This can be annoying when you don't want a `Nullable` result, e.g.:

``` julia
julia> map(isnull, NullableArray(1:3))
3-element NullableArrays.NullableArray{Any,1}:
 false
 false
 false

```

I think it would make sense to return a `NullableArray` only when the function returns `Nullable`. This is similar to operations on `BitArray`: https://github.com/JuliaLang/julia/pull/18198

davidagold:
Another way to put the issue is that the semantics for functions that act directly on the `Nullable` container, e.g. `isnull`, are different from the standard lifting semantics (as we've come to define them) for mapping arbitrary methods `f(x::T)` over `X::NullableArray{T}`. How would you feel if we defined `map(f::typeof(Base.isnull), X::NullableArray)` to give `Array` results?

Note also that comprehensions always give `Array` results, so one could use them instead:

``` julia
julia> X = NullableArray(collect(1:3))
3-element NullableArrays.NullableArray{Int64,1}:
 1
 2
 3

julia> [ isnull(x) for x in X ]
3-element Array{Bool,1}:
 false
 false
 false
```

nalimilan:
I don't really like the idea of specializing on specific functions, as it will make it very hard to understand the general behavior of `map`. As I noted in the description, I think the best behavior would be to preserve the function's return type. For cases where automatic lifting of a function is needed, people could use `@lift`. That would be much more consistent with the rest of the language.

johnmyleswhite:
To add to David's comment, the main concern I have with this proposal is that it would make it too easy for the semantics of `map` on `Array` and `NulllableArray` to diverge over time.

Consider the following example:

``` jl
f(x::Int) = 1
f(x::Nullable{Int}) = 2

map(f, 1:3)
map(f, NullableArray(1:3))
```

What do these two calls to `map` evaluate to? I would prefer that they be linked, because I think we should generally discourage redefining functions on nullable objects and instead derive their behaviors from the non-nullable definitions.

This concern is closely linked to the whitelist vs blacklist distinction I've been discussing with David for a while now:

[...]

PR #166, opened 2016-11-11
Rewrite broadcast() and map() based on lift()

nalimilan:
Remove the custom implementation of broadcast(), and just call
the base method on the lift()ed method. This implements the
blacklist approach: methods with a custom lifting behavior like
isnull() should override lift() to get passed Nullable values
directly; if they do not return a Nullable, the result is a standard
Array rather than a NullableArray.

Performance will probably be worse than before, but at least the
semantics will be correct. We can always re-implement the custom
and faster versions later when broadcast() has stabilized in Base
and Nullable support has settled.

This is WIP since I haven't thought much about this yet, because it copies `lift` from https://github.com/JuliaLang/julia/pull/18758 without too much thinking either, ~~and because it only works on 0.6 at the moment~~.

Cf. discussions at https://github.com/JuliaStats/NullableArrays.jl/issues/144 and https://github.com/JuliaStats/NullableArrays.jl/issues/163.


test/broadcast.jl
```diff
@@ -111,5 +94,52 @@ module TestBroadcast
...
+    X = NullableArray(A)
+
+    R = get.(X)
+    @test R == A
+    @test isa(R, Vector{Int})
+
+    X[1] = Nullable()
+    A[1]= 0
+    # FIXME: element-wise syntax should give the same result
+    # R = get.(X, 0)
```
nalimilan: Cf. https://github.com/JuliaLang/julia/issues/19313.

nalimilan:
@davidagold Where do you think `lift` should live? In NullableOps?

davidagold:
Preferably NullableArrays. Or am I wrong for thinking 'lift' and lifted ops
are sufficiently different strategies that they ought to live in different
packages?

nalimilan:
No, that's fine with me, I just wanted to be sure that was OK for you.

nalimilan:
https://github.com/JuliaLang/julia/issues/19313 is kind of annoying: in some cases, the `f.(x)` syntax doesn't work for two-argument functions.

There's another related issue which is even more annoying: loop fusion prevents `lift` from detecting special (blacklisted) functions. So e.g. `identity.(isnull.(x))` returns `NULL` instead of `false` for missing entries.

These two issues mostly affect corner cases, so this change is probably worth it anyway, but in the longer term I'm not sure what to do. The `f.(x)` syntax could be improved in Julia to be more flexible, i.e. calling a method with the function and input types as arguments, which could return a transformed function (lifted, in our case). Then loop fusion would combine these functions instead of the original ones. But I don't see this being accepted in Base without a strong use case, i.e. more or less official support for the lifting framework; we may as well merge https://github.com/JuliaLang/julia/pull/18758 at the same time.

I'm starting to think (again) that all function calls should lift by default; that would be a much more consistent option.

johnmyleswhite:
I'm not sure we should move forward with this just yet. It might be the right decision (and I'm inclined to think it is), but it might be easier to start with a less ambitious approach in which people have to explicitly call special functions like `isnull`. People love to complain about the burdens of those kinds of things, but I worry that this approach will be too magical and will actually prevent people from learning how to handle non-standard semantics for lifting.

nalimilan:
So that means lifting by default, without any blacklist?

bramtayl:
This kind of thing could be integrated pretty easily in ChainMap. Once a lift function is available, I could make

``` julia
@chain begin
    x
    @over is.null
    @lift_over identity
end
```

go to

`lift(broadcast, identity, broadcast(is.null, x) )`

Edit: works now on master

nalimilan:
I've updated the branch to always lift. It now uses a private `broadcast_lift` function, since for now we don't want people to override the lifting behaviour.

~~Surprisingly, new implementation appears to be just as fast as the current one. Since it passes the tests on 0.5 and 0.6, I think we can seriously consider merging it. It's not yet clear how to get non-standard lifting to work, but fixing standard cases sounds like the highest priority to me.~~
EDIT: I got the benchmarks wrong. There's still an allocation problem which kills performance. I'll investigate.

nalimilan:
Really good news: with the latest changes, I get a very similar performance to master, even slightly better, for one- and two-argument cases. Three-argument and beyond is still terribly slow; I can add a special case for 3 if needed, waiting for a more general solution.

Example on a simple benchmark:
```julia
using NullableArrays
X = NullableArray(1:100000);
Y = NullableArray(1:100000);
using BenchmarkTools

# With this branch:
julia> @benchmark broadcast!(+, X, Y)
BenchmarkTools.Trial: 
  memory estimate:  176.00 bytes
  allocs estimate:  6
  --------------
  minimum time:     107.072 μs (0.00% GC)
  median time:      108.384 μs (0.00% GC)
  mean time:        114.363 μs (0.00% GC)
  maximum time:     360.646 μs (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     1
  time tolerance:   5.00%
  memory tolerance: 1.00%

julia> @benchmark broadcast!(+, X, X, Y)
BenchmarkTools.Trial: 
  memory estimate:  224.00 bytes
  allocs estimate:  6
  --------------
  minimum time:     222.965 μs (0.00% GC)
  median time:      228.248 μs (0.00% GC)
  mean time:        246.007 μs (0.00% GC)
  maximum time:     1.621 ms (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     1
  time tolerance:   5.00%
  memory tolerance: 1.00%

# With master:
julia> @benchmark broadcast!(+, X, Y)
BenchmarkTools.Trial: 
  memory estimate:  224.00 bytes
  allocs estimate:  6
  --------------
  minimum time:     117.858 μs (0.00% GC)
  median time:      151.116 μs (0.00% GC)
  mean time:        197.762 μs (0.00% GC)
  maximum time:     1.576 ms (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     1
  time tolerance:   5.00%
  memory tolerance: 1.00%

julia> @benchmark broadcast!(+, X, X, Y)
BenchmarkTools.Trial: 
  memory estimate:  272.00 bytes
  allocs estimate:  6
  --------------
  minimum time:     243.590 μs (0.00% GC)
  median time:      251.746 μs (0.00% GC)
  mean time:        265.030 μs (0.00% GC)
  maximum time:     1.100 ms (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     1
  time tolerance:   5.00%
  memory tolerance: 1.00%

```

More generally, running `perf/broadcast.jl` shows a slight advantage to this PR, except in a handful of cases where the difference isn't that large:
https://gist.github.com/nalimilan/57ea0640ffb8fc4a8424b558efe3ff4d

Comments?

src/broadcast.jl
```diff
@@ -8,199 +9,143 @@ else
...
 end
 
-if VERSION >= v"0.5.0-dev+5189"
-    _to_shape(dims::Base.DimsOrInds) = map(_to_shape, dims)
-    _to_shape(r::Base.OneTo) = Int(last(r))
+if !isdefined(Base.Broadcast, :ftype) # Julia < 0.6
+    ftype(f, A) = typeof(f)
+    ftype(f, A...) = typeof(a -> f(a...))
+    ftype(T::DataType, A) = Type{T}
+    ftype(T::DataType, A...) = Type{T}
```
Sacha0, line 16: The changes in https://github.com/JuliaLang/julia/pull/19421 might be relevant here. Best!

src/broadcast.jl
```diff
@@ -8,199 +9,143 @@ else
...
-        return e_nullcheck
+if !isdefined(Base.Broadcast, :ziptype) # Julia < 0.6
+    if isdefined(Base, :Iterators)
+        using Base.Iterators: Zip2
+    else
+        using Base: Zip2
     end
+    ziptype(A) = Tuple{eltype(A)}
+    ziptype(A, B) = Zip2{Tuple{eltype(A)}, Tuple{eltype(B)}}
+    @inline ziptype(A, B, C, D...) = Zip{Tuple{eltype(A)}, ziptype(B, C, D...)}
```
Sacha0, line 25: Likewise here, the changes in https://github.com/JuliaLang/julia/pull/19421 might be relevant. Best!

nalimilan:
I've added the corresponding changes for `map`. Since the current implementation directly accesses the `values` and `isnull` fields, it is about twice faster as the new one. It's a bit annoying, but I'd say we should do it anyway, as we don't really have the manpower to maintain this while improving at the same time the API. With some luck, the compiler will also improve.

Calling the generic `map` implementation required changing the semantics a bit when an array is empty and no function matches the element type: we used to choose a `Union{}` eltype, while the generic implementation returns `Any`. I don't think it matters a lot anyway, but better be consistent.

EDIT: failure on 0.6 is already on master, and fixed by https://github.com/JuliaStats/NullableArrays.jl/pull/169.

REQUIRE
```diff
@@ -1,3 +1,4 @@
 julia 0.4
```
davidagold: Should we delete this line?

src/broadcast.jl
```diff
@@ -8,199 +9,92 @@ else
...
-            end
-        end
-    end
+    using Base.Broadcast: ziptype
+end
 
-    """
-        broadcast!(f, B::NullableArray, As::NullableArray...; lift::Bool=false)
+invoke_broadcast!{F, N}(f::F, dest, As::Vararg{NullableArray, N}) =
+    invoke(broadcast!, Tuple{Function, AbstractArray, Vararg{AbstractArray, N}}, f, dest, As...)
```
davidagold: Should `Function` be `F`? Probably won't make a difference in most cases.

src/broadcast.jl
```diff
@@ -8,199 +9,92 @@ else
...
+source `As` arrays as all `NullableArray`s. Thus, calling `broadcast!` on a arguments
+consisting of both `Array`s and `NullableArray`s will fall back to the standard implementation
+of `broadcast!` (i.e. without lifting).
 """
-@inline function Base.broadcast(f, Xs::NullableArray...;lift::Bool=false)
-    return broadcast!(f, NullableArray(eltype(promote_eltype(Xs...)),
-                                       _to_shape(broadcast_indices(Xs...))),
-                      Xs...; lift=lift)
+function Base.broadcast!{F, N}(f::F, dest::NullableArray, As::Vararg{NullableArray, N})
+    # These definitions are needed to avoid allocation due to splatting
```
davidagold, line 85: Have you verified that allocations are avoided for `f2` of more than two arguments? Also, did you try `@inline` decorations for `f2` to avoid the splatting penalty?

src/lift.jl
```diff
@@ -0,0 +1,61 @@
+@inline function lift(f, x)
+    if null_safe_op(f, eltype(x))
+        return @compat Nullable(f(unsafe_get(x)), !isnull(x))
+    else
+        U = Core.Inference.return_type(f, Tuple{eltype(x)})
```
davidagold: We should probably check if `U` is leaf, yes? If so, we should decide what to do when `U` is abstract. Return `Union{}` as in 16961?

src/lift.jl
```diff
@@ -0,0 +1,61 @@
...
+"""
+@inline function lift(f, xs...)
+    if null_safe_op(f, map(eltype_nullable, xs)...)
+        return @compat Nullable(f(_unsafe_get(xs...)...), !hasnulls(xs...))
+    else
+        U = Core.Inference.return_type(f, eltypes(xs...))
+        if hasnulls(xs...)
+            return Nullable{U}()
+        else
+            return Nullable(f(_unsafe_get(xs...)...))
```
davidagold: Out of curiosity, why the recursive approach as opposed to `map(unsafe_get, xs)...` ?

src/broadcast.jl
```diff
@@ -8,199 +9,92 @@ else
...
-    `NullableArray`s. Thus, calling `broadcast!` on a arguments consisting of
-    both `Array`s and `NullableArray`s will fall back to the implementation of
-    `broadcast!` in `base/broadcast.jl`.
-    """
-    # Required to solve dispatch ambiguity between
-    #   broadcast!(f, X::AbstractArray, x::Number...)
-    #   broadcast!(f, Z::NullableArrays.NullableArray, Xs::NullableArrays.NullableArray...)
-    @inline Base.broadcast!(f, Z::NullableArray; lift=false) =
-        broadcast!(f, Z, Z; lift=lift)
+Call `broadcast` with nullable lifting semantics and return a `NullableArray`.
```
davidagold, line 47: We should revise this line in light of the of the behavior defined in line 62.

src/map.jl
```diff
@@ -1,248 +1,67 @@
...
-        end
+using Base: collect_similar, Generator
+
+invoke_map!{F, N}(f::F, dest, As::Vararg{NullableArray, N}) =
+    invoke(map!, Tuple{Function, AbstractArray, Vararg{AbstractArray, N}}, f, dest, As...)
+
+"""
+    map(f, As::NullableArray...)
+
+Call `map` with nullable lifting semantics and return a `NullableArray`.
```
davidagold, line 7: Similarly should revise this description.

src/map.jl
```diff
@@ -1,248 +1,67 @@
...
-                dest.isnull[i] = true
-            else
-                dest.isnull[i] = false
-                dest.values[i] = f(X.values[j])
-            end
-        end
+using Base: collect_similar, Generator
+
+invoke_map!{F, N}(f::F, dest, As::Vararg{NullableArray, N}) =
+    invoke(map!, Tuple{Function, AbstractArray, Vararg{AbstractArray, N}}, f, dest, As...)
```
davidagold: Similarly, should we have `Function` -> `F`?

src/operators.jl
```diff
@@ -37,8 +36,7 @@ Types declared as safe can benefit from higher performance for operations on nul
 always computing the result even for null values, a branch is avoided, which helps
 vectorization.
 """
-null_safe_op(f::Any, ::Type) = false
-null_safe_op(f::Any, ::Type, ::Type) = false
+null_safe_op(f::Any, ::Type...) = false
```
davidagold, line 39: I'm confused. Isn't this in Base?

src/broadcast.jl
nalimilan: Probably, I guess it's more correct if e.g. `f` is a type and there's a specific method for them.

src/broadcast.jl
nalimilan: Good catch. Actually, since in this PR for now `lift` isn't supposed to be overloaded (so that lifting is always done), the description is correct. I've removed line 62. I hope we can improve this later in another PR.

src/broadcast.jl
nalimilan: There are no allocations on Julia master, though beyond 2 arguments performance is lower. I think I've tried everything I could (given my skills at least), without success (except a generated function, but that's probably overkill for one line). Since there are several issues open about varargs being slow, my hope is that it's going to improve at some point.

src/lift.jl
nalimilan: Makes sense.

src/lift.jl
nalimilan: In my tests it avoided allocations, which isn't the case with `map`.

src/operators.jl
nalimilan: We still carry a copy of `null_safe_op` because it doesn't exist on 0.4, and even on 0.5 it only comes with definitions for `isless` and `isequal`. But indeed we should merge them at some point (I'll do that in another PR).

nalimilan:
I've just added a commit which uses a generated function for the general `lift` method to get rid of allocations on Julia 0.5 (interestingly, the 1- and 2-argument versions do allocate there, contrary to 0.6). Unfortunately, it's still quite noticeably slower than on 0.6 (about 6x for the 1-argument version), but still "reasonable".

So I think this is now good to go if we agree on the strategy.

nalimilan:
Good news! By replacing varargs with a tuple in `lift`, I've been able to get the best performance without specialized 1- and 2-argument functions, on both Julia 0.5 and 0.6. So now only the generated function remains, and `lift` give really clean native code:
```julia
julia> @code_native NullableArrays.lift(+, (Nullable(1),Nullable(2)))
	.text
Filename: lift.jl
	pushq	%rbp
	movq	%rsp, %rbp
Source line: 21
	movq	24(%rsi), %rax
	addq	8(%rsi), %rax
Source line: 22
	movb	(%rsi), %cl
	andb	16(%rsi), %cl
Source line: 23
	andb	$1, %cl
Source line: 17
	movb	%cl, (%rdi)
	movq	%rax, 8(%rdi)
	movq	%rdi, %rax
	popq	%rbp
	retq
	nop
```

The three-argument `map` still allocates a lot currently, but this is easy to fix: https://github.com/JuliaLang/julia/pull/19522. Other than that, `map` is generally about twice slower than `broadcast`, but I'm not yet sure why. It's most likely separate from this PR anyway since both functions use the same approach.

More comments before merging?

p-zubieta:
@nalimilan Just a heads-up as I found a way of simplifying the promotion mechanism for broadcast. You can check it out here https://github.com/JuliaLang/julia/pull/19421. So you might want to wait until that lands.

nalimilan:
Yes, thanks, @Sacha0 posted the link above too. I'll update this PR once the other is merged.

nalimilan:
I've pushed a few more commits to reflect recent changes in Base. I've also included 0.6 fixes from my other PRs, unfortunately, this depends on https://github.com/JuliaLang/Compat.jl/pull/306. Finally, I've tweaked our `nullable_broadcast_eltype` to follow https://github.com/JuliaLang/julia/pull/20049, since we use the same code for `broadcast` and `map` (could have to change if that PR isn't merged).

So I think it's good to go after the new Compat release. Any more comments?

davidagold:
None from me, I don't think. I'm very excited about this.

nalimilan:
Should be good to go now, the 0.6 failure is in operators.jl.

src/broadcast.jl
```diff
@@ -25,7 +25,7 @@ if VERSION < v"0.6.0-dev" # Old approach needed for inference to work
     @inline ziptype(A, B, C, D...) = Zip{Tuple{eltype(A)}, ziptype(B, C, D...)}
 
     nullable_broadcast_eltype(f, As...) =
-        _default_eltype(Base.Generator{ziptype(As...), ftype(f, As...)})
+        eltype(_default_eltype(Base.Generator{ziptype(As...), ftype(f, As...)}))
```
p-zubieta, line 28: I'm curious why is this needed?

nalimilan: What is "this"?

p-zubieta: Sorry, the `eltype` you last added.

nalimilan: It's useful because `NullableArray{T}` is parameterized on the eltype of `Nullable` rather than on the `Nullable{T}` itself.

p-zubieta: But, wouldn't that warrants that you don't get a `Nullable{T}` here?

Given that `eltype(NullableArray([1], [true])) == Int`, that implies `_default_eltype(+, NullableArray([1], [true]), NullableArray([1], [true])) == Int`, right?

nalimilan: No, `eltype(NullableArray([1], [true])) == Nullable{Int}`.

I don't want `T = Nullable{Int}` here since it's used to build a new array via `NullableArray{T}(...)`. `NullableArray{Nullable{Int}}` would have a `Nullable{Nullable{Int}}` eltype.

p-zubieta: The problem with this is that `eltype(_default_eltype(*, ["a"], ["b"])` will give `Char`, so I guess you'd need a custom `eltype` here.

nalimilan: But that method is only called on `NullableArrays`, so we're certain the element types are `Nullable` (in your case, `Nullable{String}`).

p-zubieta: Oh, now I see, this is `nullable_broadcast_eltype` as opposed to `broadcast_eltype`. Sorry for the noise then.

nalimilan:
Will squash and merge tomorrow unless somebody objects or beats me to it.

test/map.jl
```diff
@@ -83,79 +65,59 @@ module TestMap
...
-    h2(x::Nullable...) = prod(x)
+    h2(x) = x
+    h2(x...) = prod(x)
 
     Z1 = map(h1, X)
     Z2 = map(h2, X)
-    if VERSION >= v"0.5.0-dev+3294"
-        @test isequal(Z1, NullableArray(Union{}[])) # no type inference
+    @test isempty(Z1)
+    if VERSION >= v"0.6.0-dev"
```
tkelman, line 83: please be specific with VERSION cutoffs like this, will this really work for all 0.6-dev versions?

nalimilan: Actually, looks like I fixed this in https://github.com/JuliaStats/NullableArrays.jl/pull/177/files. Will remove the `if`.

(FWIW, at the time I added these conditions, the issue was that inference failed on 0.5.0, and I had no idea which commit(s) fixed it except for bisecting it...)

nalimilan: See https://github.com/JuliaStats/NullableArrays.jl/pull/183.

PR #168, opened 2016-11-25
Fix deprecation of $ as xor()

nalimilan:


ararslan (approved):
Who knows what's going on with AppVeyor. This looks good to me, and Travis seems to agree.

nalimilan:
There's something weird with AppVeyor, it always starts twice, and the first build fails because a newer one is queued. When the second build is done, it turns green. Never seen this in other projects, so the configuration must be incorrect.

nalimilan:
Turns out the webhook was set up twice...

ararslan:
Weird. Were you able to fix that?

PR #170, opened 2016-12-27
WIP: allow value array to be any AbstractArray

shashi:
This makes possible construction of NullableArray of StructOfArrays, or other [funky arrays](https://github.com/JuliaData/CSV.jl/blob/speedups/src/textparser/lib/substringarray.jl#L6-L9) which solve orthogonal problems.

The following fails with this patch applied:

```julia
julia> isequal(NullableArray(Union{},1), NullableArray(Float64,1))
ERROR: indexing not defined for NullableArrays.NullableArray{Union{},1}
 in _getindex(::Base.LinearFast, ::NullableArrays.NullableArray{Union{},1}, ::Int64) at ./abstractarray.jl:763
 in next at ./iterator.jl:98 [inlined]
 in isequal(::NullableArrays.NullableArray{Union{},1}, ::NullableArrays.NullableArray{Float64,1}) at ./abstractarray.jl:1385
```
Stepping through this with Gallium doesn't fail with this error, it instead succeeds in returning `true`! cc @keno

### DIFF
--- a/src/typedefs.jl
+++ b/src/typedefs.jl
@@ -19,7 +19,7 @@ It allows users to easily define operations on arrays with null values by
 reusing operations that only work on arrays without any null values.
 """
 immutable NullableArray{T, N} <: AbstractArray{Nullable{T}, N}
-    values::Array{T, N}
+    values::AbstractArray{T, N}
     isnull::Array{Bool, N}
     # extra field for potentially holding a reference to a parent memory block
     # (think mmapped file, for example) that `values` is actually derived from
--- a/src/constructors.jl
+++ b/src/constructors.jl
@@ -83,8 +83,7 @@ end
 #
 function NullableArray{T}(A::AbstractArray,
                              ::Type{T},
-                             na::Any;
-                             conversion::Base.Callable=Base.convert) # -> NullableArray{T, N}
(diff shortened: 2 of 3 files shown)


shashi:
Note that there are no problems if we are comparing nullablearrays of other types:

```
julia> isequal(NullableArray(Int64,1), NullableArray(Float64,1))
true
```

This strikes me as strange (it is possible I'm doing something really stupid). I suspect the problem is `Union{}` and inlining but I don't know how to dive deeper into this...

shashi:
Super weird: test passes on 0.4 and 0.5 on travis but appveyor fails with the same error

image

src/typedefs.jl
```diff
@@ -19,7 +19,7 @@ It allows users to easily define operations on arrays with null values by
 reusing operations that only work on arrays without any null values.
 """
 immutable NullableArray{T, N} <: AbstractArray{Nullable{T}, N}
-    values::Array{T, N}
+    values::AbstractArray{T, N}
```
nalimilan, line 22: This should be made a type parameter of the array or it's going to kill performance.

nalimilan:
Just bumped into this, might be related to the bug you're getting? https://github.com/JuliaLang/julia/issues/10326

PR #171, opened 2017-01-14
0.6 fixes

nalimilan:
Three fixes towards having tests passing again on Julia 0.6. There remains a failure related to `map`, which will be fixed by https://github.com/JuliaStats/NullableArrays.jl/pull/166.

### DIFF
--- a/src/broadcast.jl
+++ b/src/broadcast.jl
@@ -204,15 +204,17 @@ implementation of `broadcast` in `base/broadcast.jl`.
 end
 
 # broadcasted ops
-for (op, scalar_op) in (
-    (:(@compat Base.:(.==)), :(==)),
-    (:(@compat Base.:.!=), :!=),
-    (:(@compat Base.:.<), :<),
-    (:(@compat Base.:.>), :>),
-    (:(@compat Base.:.<=), :<=),
-    (:(@compat Base.:.>=), :>=)
-)
-    @eval begin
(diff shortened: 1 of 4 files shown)


davidagold:
Looks good to me. @ararslan ?

src/broadcast.jl
```diff
@@ -204,15 +204,17 @@ implementation of `broadcast` in `base/broadcast.jl`.
...
+    for (op, scalar_op) in (
+        (:(@compat Base.:(.==)), :(==)),
+        (:(@compat Base.:.!=), :!=),
+        (:(@compat Base.:.<), :<),
+        (:(@compat Base.:.>), :>),
+        (:(@compat Base.:.<=), :<=),
+        (:(@compat Base.:.>=), :>=)
+    )
+        @eval begin
+            ($op)(X::NullableArray, Y::NullableArray) = broadcast($scalar_op, X, Y)
```
ararslan, line 217: Isn't this what's causing the warnings on nightly? For example,
```
WARNING: .* is no longer a function object; use broadcast(*, ...) instead
```

nalimilan: No, it actually comes from the tests which loop other these functions, which is deprecated. I take care of that in the other PR.

ararslan: Okay cool, thanks for the explanation!

Issue #174, opened 2017-01-24
broadcast with arguments of different type


piever:
First of all, kudos for #166, I find it extremely useful.
I have one doubt: by looking at the implementation, it seems that it only works if all the arguments of the function have the same type, i.e. this works fine:

`NullableArray(ones(100)) .+ NullableArray(ones(100))`

whereas this throws an error:

`NullableArray(ones(Int64, 100)) .+ NullableArray(ones(Float64,100))`

which as far as I understand comes from [this definition of lift](https://github.com/JuliaStats/NullableArrays.jl/blob/master/src/lift.jl#L16).

Is there a way/plan to generalize lift to tuples with elements of different types, or is that unfeasible?

davidagold:
Do you mean to say that `NullableArray(ones(100)) .+ ones(100)` throws an error?

piever:
Sorry! I copied the same code twice, I've edited the post above. Running this:

`NullableArray(ones(Int64, 100)) .+ NullableArray(ones(Float64,100))`

gives the following error:

`ERROR: MethodError: no method matching lift(::Base.#+, ::Tuple{Nullable{Int64},Nullable{Float64}})`

Of course this example is not useful in practice, but there are several use-cases where one would want to broadcast a function taking arguments of different types.

EDIT: of course you're right, I've changed it again, this is what give the `lift` related error message.

davidagold:
Well, the trouble there is that you have reversed the arguments to `ones`. But in general you seem to be right:

```julia
julia> NullableArray(ones(Int64, 100)) .+ NullableArray(ones(Float64, 100))
WARNING: Array{T,N}(::Type{T},d::NTuple{N,Int}) is deprecated, use Array{T,N}(d) instead.
Stacktrace:
 [1] depwarn(::String, ::Symbol) at ./deprecated.jl:64
 [2] Array{T,N}(::Type{Float64}, ::Tuple{Int64}) at ./deprecated.jl:50
 [3] Type at /Users/David/.julia/v0.6/NullableArrays/src/constructors.jl:25 [inlined]
 [4] Type at /Users/David/.julia/v0.6/NullableArrays/src/constructors.jl:35 [inlined]
 [5] similar at ./abstractarray.jl:489 [inlined]
[...]

PR #175, opened 2017-01-24
Fix broadcast() and map() when passed arrays with different eltypes

nalimilan:
The explicit T type parameter to lift() forced input values to be of the
same type. Turns out it is not needed to force specialization.
Also add tests for this.

Fixes https://github.com/JuliaStats/NullableArrays.jl/issues/174


davidagold (approved):

davidagold:
that depwarn tho

davidagold:
The nightly failure looks unrelated. Is it new?

nalimilan:
No, it's been there since the type system changes I think. Anyway AFAIK we should remove operators.jl on 0.6 since we can use the element-wise operators now.

Issue #176, opened 2017-01-25
Broadcast not working on type constructors with NullableArrays (on master)


piever:
After #175 I can pass nullable arrays with different `eltype` to `broadcast` but it still doesn't seem to work on type constructors.

Example:

```
type SurvEvent
    time
    censored
end
t = NullableArray(rand(3))
c = NullableArray(rand(Bool,3))
SurvEvent.(t,c)

UndefRefError: access to undefined reference

 in lift(...) at /Users/pietro/.julia/v0.5/NullableArrays/src/lift.jl:20
 in macro expansion at ./broadcast.jl:129 [inlined]
 in macro expansion at ./simdloop.jl:73 [inlined]
 in macro expansion at ./broadcast.jl:123 [inlined]
 in _broadcast!(::NullableArrays.#f2#17{DataType}, ::NullableArrays.NullableArray{Any,1}, ::Tuple{Tuple{Bool},Tuple{Bool}}, ::Tuple{Tuple{Int64},Tuple{Int64}}, ::Tuple{NullableArrays.NullableArray{Float64,1},NullableArrays.NullableArray{Bool,1}}, ::Type{Val{2}}) at ./broadcast.jl:117
 in broadcast!(::Function, ::NullableArrays.NullableArray{Any,1}, ::NullableArrays.NullableArray{Float64,1}, ::NullableArrays.NullableArray{Bool,1}) at ./broadcast.jl:172
 in broadcast(::Type{SurvEvent}, ::NullableArrays.NullableArray{Float64,1}, ::NullableArrays.NullableArray{Bool,1}) at /Users/pietro/.julia/v0.5/NullableArrays/src/broadcast.jl:69
```

On the other hand, this works (but it fails the inference part as the output should be of type `NullableArrays.NullableArray{SurvEvent,1}` ):

```
g(x,y) = SurvEvent(x,y)
g.(t,c)

3-element NullableArrays.NullableArray{Any,1}:
 SurvEvent(0.914069,true) 
 SurvEvent(0.702524,false)
 SurvEvent(0.840124,true) 
```

nalimilan:
Indeed, the `f.instance` on that line wasn't a very clean approach. I guess that's what happens when you do something without knowing why it works...

I have a simple fix, but I need to check that performance doesn't suffer first.

nalimilan:
[...]

PR #177, opened 2017-01-25
Fix broadcast() and map() with constructors

nalimilan:
Non-singleton types do not have an 'instance' field, so we cannot call
null_safe_op() from a generated function for them. This means they will
always be considered as unsafe.

Unfortunately, this means that the return type cannot be computed from
the generated function either, and type inference only works on Julia 0.6.

Fixes https://github.com/JuliaStats/NullableArrays.jl/issues/176.

### DIFF
--- a/src/broadcast.jl
+++ b/src/broadcast.jl
@@ -1,4 +1,3 @@
-using Base: _default_eltype
 using Compat
 
 if VERSION >= v"0.6.0-dev.693"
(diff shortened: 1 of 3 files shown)


nalimilan:
I've found a bug in the existing code on Julia 0.5, and I figured an ugly way of computing the return type on Julia 0.5. The new code should work as expected everywhere (even if constructors cannot be marked as `null_safe_op`).

Issue #167, opened 2016-11-16
Override vcat() and (maybe) promote_rule()


nalimilan:
We need to override `vcat` to fix this inconsistency:
```julia
julia> vcat(NullableArray(1:2), 1:3)
5-element NullableArrays.NullableArray{Int64,1}:
 1
 2
 1
 2
 3

julia> vcat(1:2, NullableArray(1:2))
4-element Array{Nullable{Int64},1}:
 1
 2
 1
 2

```

The first case is OK, the second one isn't.

By the way, it could make sense to override `promote_rule` too, since ideally `vcat` would use it. Cf. https://github.com/JuliaData/CategoricalArrays.jl/pull/42.

cjprybol:
issue is with `typed_vcat` in Base always using the first `vcat` element to call `similar`.

```julia
julia> similar(full(NullableArray(1:2)), Nullable{Int64}, 4)
4-element NullableArrays.NullableArray{Int64,1}:
 #NULL
 #NULL
 #NULL
 #NULL

julia> similar(full(1:2), Nullable{Int64}, 4)
4-element Array{Nullable{Int64},1}:
 4626872848
 4626872880
 4626872880
 4639814992
```

See `typed_vcat` function defintions for [`AbstractVector`](https://github.com/JuliaLang/julia/blob/30a215d19aab3adff829b3af90e2d5c7ade978bd/base/abstractarray.jl#L1046) & [`AbstractMatrix`](https://github.com/JuliaLang/julia/blob/30a215d19aab3adff829b3af90e2d5c7ade978bd/base/abstractarray.jl#L1106). We can fix it by asking `typed_vcat` to check if any `vcat` inputs are `NullableArrays` and, if so, call `similar` with that input instead of the first.

```julia
# fixed
julia> vcat(1:2, NullableArray(1:2))
4-element NullableArrays.NullableArray{Int64,1}:
 1
 2
 1
 2
```

We need the same change for hcat
```julia
# this is wrong
julia> hcat(1:2, NullableArray(1:2))
2×2 Array{Nullable{Int64},2}:
 1  1
 2  2

julia> hcat(NullableArray(1:2), 1:2)
2×2 NullableArrays.NullableArray{Int64,2}:
 1  1
 2  2
```

And this `promote_rule` seems wrong? Is `promote_rule` supposed to be reciprocal?
```julia
julia> promote_rule(Nullable{Int64}, Int64)
Nullable{Int64}

[...]

PR #187, opened 2017-02-24
address vcat return inconsistency

cjprybol:
https://github.com/JuliaStats/NullableArrays.jl/issues/167


nalimilan:
Thanks, but I'd really like to avoid having to copy so much code from Base. This is https://github.com/JuliaLang/julia/issues/2326.

In the short term, I'm not sure what's the best approach. Isn't there a way to use the machinery from Base without reimplementing the full method?

src/nullablevector.jl
```diff
@@ -311,3 +311,80 @@ function Base.empty!(X::NullableVector)
     empty!(X.isnull)
     return X
 end
+
+function Base.promote_rule{T1,T2}(::Type{T1}, ::Type{Nullable{T2}})
```
nalimilan: This is not actually needed (and would have to live in Base anyway). Indeed you should never call `promote_rule` directly, but use `promote_type` instead. This already works:
```julia
julia> promote_type(Nullable{Int64}, Int64)
Nullable{Int64}
```

nalimilan:
I guess one partial solution is to define methods like `vcat(::AbstractArray, ::NullableArray, ::AbstractArray...)` which call `Base.type_hcat` with the correct return type. Cf. https://github.com/JuliaStats/DataArrays.jl/pull/213.

src/nullablevector.jl
```diff
@@ -311,3 +311,80 @@ function Base.empty!(X::NullableVector)
...
+    promote_rule(Nullable{T2}, T1)
+end
+
+function Base.typed_hcat{T}(::Type{T}, A::AbstractVecOrMat...)
```
nalimilan: It's not OK to override Base methods like this with signatures that only involve Base types. That will affect unrelated packages and is known as type piracy.

src/nullablevector.jl
cjprybol: Yes, you're right. I see now that `promote_type` has the consistency I was expecting from `promote_rule`. I'll need to learn more about those differences.

```julia
julia> promote_rule(Nullable{Int64}, Int64)
Nullable{Int64}

julia> promote_rule(Int64, Nullable{Int64})
Union{}

julia> promote_type(Nullable{Int64}, Int64)
Nullable{Int64}

julia> promote_type(Int64, Nullable{Int64})
Nullable{Int64}
```

src/nullablevector.jl
cjprybol: That makes perfect sense. Thanks for explaining what type piracy is

cjprybol:
I'm not sure how to do this without overwriting the functions. I tried several combinations of extending `hcat` and `vcat` for when the second input is a `NullableArray` but the calls kept utilizing the broad definitions in base. Am I missing something?

```julia
using Gallium
using NullableArrays
import Base.hcat

julia> @enter hcat(1:2, NullableArray(3:4))
In abstractarray.jl:1001
1000
1001    hcat(A::AbstractVecOrMat...) = typed_hcat(promote_eltype(A...), A...)
1002    hcat{T}(A::AbstractVecOrMat{T}...) = typed_hcat(T, A...)
1003

About to run: (Core._apply)(Base.promote_eltype,(1:2,Nullable{Int64}[3,4]))
1|debug >

julia> # try to define a more specific function definition

julia> hcat{T<:NullableArray}(A, B::T) = typed_hcat(promote_eltype(A...), A...)
hcat (generic function with 22 methods)

julia> # still calls the same function

julia> @enter hcat(1:2, NullableArray(3:4))
In abstractarray.jl:1001
1000
1001    hcat(A::AbstractVecOrMat...) = typed_hcat(promote_eltype(A...), A...)
1002    hcat{T}(A::AbstractVecOrMat{T}...) = typed_hcat(T, A...)
1003

About to run: (Core._apply)(Base.promote_eltype,(1:2,Nullable{Int64}[3,4]))
```

cjprybol:
I need to figure out how to handle the 3+ argument case but this is working nicely for the 2 argument cases

```julia
julia> @test typeof(vcat(1:2, NullableArray(3:4))) == NullableArrays.NullableArray{Int64,1}
Test Passed
  Expression: typeof(vcat(1:2,NullableArray(3:4))) == NullableArrays.NullableArray{Int64,1}
   Evaluated: NullableArrays.NullableArray{Int64,1} == NullableArrays.NullableArray{Int64,1}

julia> @test typeof(vcat([1 2], NullableArray([3 4]))) == NullableArrays.NullableArray{Int64,2}
Test Passed
  Expression: typeof(vcat([1 2],NullableArray([3 4]))) == NullableArrays.NullableArray{Int64,2}
   Evaluated: NullableArrays.NullableArray{Int64,2} == NullableArrays.NullableArray{Int64,2}

julia> @test typeof(hcat(1:2, NullableArray(3:4))) == NullableArrays.NullableArray{Int64,2}
Test Passed
  Expression: typeof(hcat(1:2,NullableArray(3:4))) == NullableArrays.NullableArray{Int64,2}
   Evaluated: NullableArrays.NullableArray{Int64,2} == NullableArrays.NullableArray{Int64,2}

julia> @test typeof(hcat([1 2], NullableArray([3 4]))) == NullableArrays.NullableArray{Int64,2}
Test Passed
  Expression: typeof(hcat([1 2],NullableArray([3 4]))) == NullableArrays.NullableArray{Int64,2}
   Evaluated: NullableArrays.NullableArray{Int64,2} == NullableArrays.NullableArray{Int64,2}
```

src/nullablevector.jl
```diff
@@ -311,3 +311,14 @@ function Base.empty!(X::NullableVector)
     empty!(X.isnull)
     return X
 end
+
+notnullarray = Union{filter!(x -> !isa(x, Type{NullableArray}), subtypes(AbstractArray))...}
```
nalimilan: You shouldn't need this. Just `AbstractArray` should be enough, and the most specific method will take precedence.

src/nullablevector.jl
```diff
@@ -311,3 +311,14 @@ function Base.empty!(X::NullableVector)
...
+notnullmatrix = typeintersect(notnullarray, AbstractMatrix)
+notnullvector = typeintersect(notnullarray, AbstractVector)
+
+Base.vcat{T <: notnullarray}(A::T, B::NullableArray) = Base.vcat(NullableArray(A), B)
```
nalimilan: This is wasteful since it creates a copy. Maybe we can use this trick to force creating a `NullableArray`: `vcat(NullableArray{eltype(A)}(), A, B)`.

nalimilan: Also, no need for `Base.`.

src/operators.jl
```diff
@@ -1,7 +1,7 @@
 ## Lifted operators
 
 importall Base.Operators
-import Base: promote_op, abs, abs2, sqrt, cbrt, scalarmin, scalarmax, isless
+import Base: abs, abs2, cbrt, isless, promote_op, scalarmin, scalarmax, sqrt
```
nalimilan: Unrelated.

test/nullablevector.jl
```diff
@@ -223,4 +223,21 @@ module TestNullableVector
     X = NullableArray(A, M)
     empty!(X)
     @test isempty(X)
+
+    @test typeof(hcat(NullableArray(1:2), 3:4)) == NullableArrays.NullableArray{Int64,2}
```
nalimilan: No need for `NullableArrays.`.

src/nullablevector.jl
```diff
@@ -1,3 +1,5 @@
+import Base: hcat, vcat
```
cjprybol: This file uses `Base.` to prefix the function extensions. Should I add all of the functions to this import call? It's an unrelated change but importing these two and leaving everything else to be `Base.` is inconsistent

src/nullablevector.jl
cjprybol: When using`AbstractArray` there's no way to exit the `vcat -> vcat` loop and it results in StackOverflow. We either eternally prepend empty NullableArrays (`vcat(NullableVector{eltype(A)}(), A, B)`) or eternally try to convert A to a `NullableArray` (`vcat(NullableArray(A), B)`)

src/nullablevector.jl
cjprybol: I think this case of adding an empty NullableArray to the beginning of the `vcat/hcat` call only works on `vcat` with `Vector`s, but I've added it for that case. Every other case I tried will fail in dimension mismatch checks

src/operators.jl
cjprybol: reverted

test/nullablevector.jl
cjprybol: removed

src/nullablevector.jl
cjprybol: I now realize that if I call `typed_hcat`, `typed_vcat` instead of calling `hcat/vcat` I can probably avoid the StackOverflow while using the simpler `AbstractArray` as you suggested.

cjprybol:
Should these promote to NullableCategoricalArray?
```julia
julia> vcat(categorical([1,2,3]), NullableArray([4,5,6]))
6-element NullableArrays.NullableArray{Int64,1}:
 1
 2
 3
 4
 5
 6

julia> vcat(NullableArray([4,5,6]), categorical([1,2,3]))
6-element NullableArrays.NullableArray{Int64,1}:
 4
 5
 6
 1
 2
 3
```

src/nullablevector.jl
```diff
@@ -311,3 +313,27 @@ function Base.empty!(X::NullableVector)
     empty!(X.isnull)
     return X
 end
+
+function vcat(A::AbstractVector, B::AbstractVector...)
```
nalimilan: This is still type piracy. There needs to be `NullableArray` in the signature. Currently I don't think this can be fixed in full generality. We can include a few methods for all possible combinations up to e.g. 3 arguments, i.e:
- `::NullableArray, ::AbstractArray`
- `::AbstractArray, ::NullableArray`
- `::NullableArray, ::AbstractArray, ::AbstractArray`
- `::AbstractArray, ::NullableArray, ::AbstractArray`
- `::AbstractArray, ::AbstractArray, ::NullableArray`

src/nullablevector.jl
nalimilan: Yeah, better not import anything and use the `Base.` prefix when overloading methods.

nalimilan:
> Should these promote to NullableCategoricalArray?

Yes, but that's tricky since doing so forces one package to depend on the other one, and as long as StatsModels depends on CategoricalArrays we'd better get rid of the NullableArrays dependency (https://github.com/JuliaData/CategoricalArrays.jl/pull/55). So for now I would leave this case as is.

src/nullablevector.jl
cjprybol: I think defining these functions would be a regression compared to just leaving the current functionality as is.
```julia
# this rule seems more intuitive
isa(array[1], NullableArray) ? NullableArray{T} : Array[Nullable{T}]

# than this rule
isa(array[1], NullableArray) ||
isa(array[2], NullableArray) ||
isa(array[3], NullableArray) ? NullableArray{T} : Array[Nullable{T}]
```

If the general functionality is type piracy then I feel the best way forward would be to either address https://github.com/JuliaLang/julia/issues/2326 directly or accept the current behavior. Thoughts?

nalimilan: I really don't know. Clearly we can't reach a satisfying state without improvements in Julia. But which of the partial solutions is better, I don't know...

cjprybol: Similar sentiments here. I'll implement the 2 argument cases for `hcat/vcat` and now at least we have everything written up and documented.

src/nullablevector.jl
```diff
@@ -314,26 +314,19 @@ function Base.empty!(X::NullableVector)
     return X
 end
 
-function vcat(A::AbstractVector, B::AbstractVector...)
-    any(b -> isa(b, NullableArray), B) ?
-        typed_vcat(promote_eltype(A, B...), NullableArray(A), B...) :
-        typed_vcat(promote_eltype(A, B...), A, B...)
+function vcat(A::AbstractVector, B::NullableVector)
+        typed_vcat(promote_eltype(A, B), NullableArray(A), B)
```
ararslan: Please reduce the number of spaces used for indentation here by 50%

src/nullablevector.jl
nalimilan: If you implement the two-argument versions, please also implement the three-argument versions. Else the rule is going to be even more complex than what you wrote above.

Of course, this rests on the assumption that concatenating more than 3 arrays is a somewhat rare operation, so that it's not a big issue if other variants behave differently. Might not be the case, but only special-casing two-argument methods is clearly a worse solution.

cjprybol:
I think we should close this. I read through a few dozen pages of search results for `vcat` usage in Julia code here on GitHub and saw a usage pattern ~ `vcat(a,b) >> vcat(a...) ~= vcat(a,b,c) >> vcat(a,b,c,d)`. If we could redirect this into a brainstorm about how to address https://github.com/JuliaLang/julia/issues/2326, I think that would be a much better use of our time and I'd be happy to try and implement suggestions. I tried variants of https://github.com/JuliaLang/julia/issues/2326#issuecomment-44790021 and couldn't get any of them to work

nalimilan:
I've tried starting a discussion about the way to move forward at https://github.com/JuliaLang/julia/pull/20815, let's see what can be done about this.

ararslan:
In the meantime, thanks so much for your hard work here, Cameron. Much appreciated.

PR #188, opened 2017-03-23
short circuit anynull if eltype !<: Nullable

cjprybol:
example
```julia
using NullableArrays
using BenchmarkTools
X = rand(Int(1e6));
Y = rand(Int(1e4), Int(1e4));
@benchmark anynull(X)
@benchmark anynull(Y)
```

before
```julia
julia> @benchmark anynull(X)
BenchmarkTools.Trial:
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     270.964 μs (0.00% GC)
  median time:      396.844 μs (0.00% GC)
  mean time:        373.259 μs (0.00% GC)
  maximum time:     917.353 μs (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     1
  time tolerance:   5.00%
  memory tolerance: 1.00%

julia> @benchmark anynull(Y)
BenchmarkTools.Trial:
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     27.749 ms (0.00% GC)
  median time:      29.160 ms (0.00% GC)
  mean time:        29.763 ms (0.00% GC)
  maximum time:     36.749 ms (0.00% GC)
  --------------
  samples:          168
  evals/sample:     1
  time tolerance:   5.00%
  memory tolerance: 1.00%
```

after
```julia
julia> @benchmark anynull(X)
BenchmarkTools.Trial:
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     18.482 ns (0.00% GC)
  median time:      19.022 ns (0.00% GC)
  mean time:        20.409 ns (0.00% GC)
  maximum time:     87.697 ns (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     997
  time tolerance:   5.00%
  memory tolerance: 1.00%

julia> @benchmark anynull(Y)
BenchmarkTools.Trial:
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     17.400 ns (0.00% GC)
  median time:      17.926 ns (0.00% GC)
  mean time:        19.375 ns (0.00% GC)
  maximum time:     82.290 ns (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     998
  time tolerance:   5.00%
  memory tolerance: 1.00%
```


src/primitives.jl
```diff
@@ -210,8 +210,9 @@ dropnull!(X::NullableVector) = deleteat!(X, find(X.isnull)).values # -> Vector
 
 Returns whether or not any entries of `X` are null.
 """
-anynull(X::Any) = any(_isnull, X)           # -> Bool
-anynull(X::NullableArray) = any(X.isnull)   # -> Bool
+anynull(X::Any) = any(_isnull, X)                                           # -> Bool
+anynull{T}(X::AbstractArray{T}) = T <: Nullable ? any(_isnull, X) : false   # -> Bool
```
ararslan: You should be able to avoid the branch entirely and use dispatch here.
```julia
anynull{T<:Nullable}(X::AbstractArray{T}) = any(_isnull, X)
anynull(X::AbstractArray) = false
```

cjprybol:
I'm not sure what the coverage drop is for in regards to coveralls, but I could open another PR to try and extend test coverage?

ararslan:
Weird that Codecov went up negligibly and Coveralls fell significantly... not sure how to interpret that. I guess I'm inclined to call it a fluke, but I'll leave that up to @nalimilan.

ararslan (approved):

src/primitives.jl
nalimilan: You also need to check `T :> Nullable` too, so you could as well keep checks inside the function.

nalimilan (changes requested):
I wouldn't trust the coverage measures too much, but you probably need new tests to cover non-nullable arrays (and `Array{Any}` after my last comment).

src/primitives.jl
```diff
@@ -210,8 +210,18 @@ dropnull!(X::NullableVector) = deleteat!(X, find(X.isnull)).values # -> Vector
 
 Returns whether or not any entries of `X` are null.
 """
-anynull(X::Any) = any(_isnull, X)           # -> Bool
-anynull(X::NullableArray) = any(X.isnull)   # -> Bool
+anynull(X::Any) = any(_isnull, X)                              # -> Bool
+anynull(X::NullableArray) = any(X.isnull)                      # -> Bool
+function anynull{T}(X::AbstractArray{T})                       # -> Bool
+    u = isa(T, Union)
+    if u && !any(f -> eval(:($T.$f)) <: Nullable || Nullable <: eval(:($T.$f)), fieldnames(T))
```
cjprybol: We all missed this case when implementing the short-circuit for dropnull & dropnull!, but Unions are tricky to handle. Any ideas how to do this without the eval? I couldn't figure out how to iterate over the types of the union without it.

ararslan:
```julia
!any(x -> (S = fieldtype(T, x); S <: Nullable || Nullable <: S), fieldnames(T))
```

ararslan: Also, 0.5:
```julia
!any(S -> S <: Nullable || Nullable <: S, T.types)
```
0.6:
```julia
!any(S -> S <: Nullable || Nullable <: S, Base.uniontypes(T))
```
Unfortunately neither of those nicer ones work on both versions.

ararslan: Nevermind, the first thing I said won't work at all; I think you need to condition on the version and do the second thing

src/primitives.jl
```diff
@@ -214,13 +214,16 @@ anynull(X::Any) = any(_isnull, X)                              # -> Bool
...
     u = isa(T, Union)
-    if u && !any(f -> eval(:($T.$f)) <: Nullable || Nullable <: eval(:($T.$f)), fieldnames(T))
+    if !u && !(Nullable <: T) && !(T <: Nullable)
         return false
-    elseif !u && !(Nullable <: T) && !(T <: Nullable)
-        return false
-    else
-        return any(_isnull, X)
+    if isa(T, Union)
+        if VERSION < v"0.6.0-" && !any(S -> S <: Nullable || Nullable <: S, T.types)
```
ararslan: `@static if`

cjprybol: `@static` doesn't seem to be parsing correctly on 0.5. Do you mind if I revert that change? Or is there another way I should write it?

ararslan: Ah yeah, part of that is not a statically evaluable condition, so you'd have to do nested `if`s.

Instead, at the top level outside of this function
```julia
if VERSION < v"0.6.0-dev.2107"
    _uniontypes(T::Union) = T.types
else
    _uniontypes(T::Union) = Base.uniontypes(T)
end
```
Then use `_uniontypes(T)` in the function here.

cjprybol: Great suggestion, thanks!

cjprybol: and I also noticed you tracked down the specific version where that change was made 🥇

cjprybol:
All tests but coveralls are happy again. If @nalimilan is ok with these changes when the sun comes up on his side of the world, then we can merge and I'll followup with another PR to make the same changes for dropnull & dropnull! and add appropriate tests. Thanks for the help and reviews to both of you!

PR #191, opened 2017-04-13
Move StatsBase describe code from DataTables to NullableArrays

cjprybol:
Also makes minor adjustments from current DataTables format to make printing more like StatsBase. Follows https://github.com/JuliaStats/StatsBase.jl/pull/247.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,3 +1,4 @@
 julia 0.5
 Compat 0.19.0
 Reexport
+StatsBase 0.14
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -1,0 +1,53 @@
+StatsBase.describe(X::NullableVector) = StatsBase.describe(STDOUT, X)
+
+function StatsBase.describe{T<:Real}(io::IO, X::NullableVector{T})
+    nullcount = sum(X.isnull)
+    pnull = 100nullcount/length(X)
+    if pnull != 100 # describe will fail if dropnull returns an empty vector
+        describe(io, dropnull(X))
+    else
+        println(io, "Summary Stats:")
+        println(io, "Type:           $(eltype(X))")
+    end
+    println(io, "Number Missing: $(nullcount)")
+    @printf(io, "%% Missing:      %.6f\n", pnull)
+    return
+end
+
+function StatsBase.describe(io::IO, X::NullableVector)
+    nullcount = sum(X.isnull)
+    pnull = 100nullcount/length(X)
+    println(io, "Summary Stats:")
+    println(io, "Length:         $(length(X))")
+    println(io, "Type:           $(eltype(X))")
+    println(io, "Number Unique:  $(length(unique(X)))")
+    println(io, "Number Missing: $(nullcount)")
+    @printf(io, "%% Missing:      %.6f\n", pnull)
+    return
+end
+
+function StatsBase.describe{T<:Real}(io::IO, X::AbstractVector{Nullable{T}})
+    nullcount = sum(_isnull, X)
+    pnull = 100nullcount/length(X)
+    if pnull != 100 # describe will fail if dropnull returns an empty vector
+        describe(io, dropnull(X))
+    else
+        println(io, "Summary Stats:")
+        println(io, "Type:           $(eltype(X))")
+    end
+    println(io, "Number Missing: $(nullcount)")
+    @printf(io, "%% Missing:      %.6f\n", pnull)
+    return
+end
+
+function StatsBase.describe{T<:Nullable}(io::IO, X::AbstractVector{T})
+    nullcount = sum(_isnull, X)
+    pnull = 100nullcount/length(X)
+    println(io, "Summary Stats:")
+    println(io, "Length:         $(length(X))")
+    println(io, "Type:           $(eltype(X))")
+    println(io, "Number Unique:  $(length(unique(X)))")
+    println(io, "Number Missing: $(nullcount)")
+    @printf(io, "%% Missing:      %.6f\n", pnull)
+    return
+end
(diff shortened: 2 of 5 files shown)


src/utils.jl
```diff
@@ -0,0 +1,49 @@
+StatsBase.describe(X::NullableVector) = StatsBase.describe(STDOUT, X)
+
+function StatsBase.describe(io::IO, X::NullableVector{<:Real})
+    if allnull(X)
+        println(io, " * All null * ") && return
```
ararslan: This will fail since `println` returns `Void` and it's being used in a boolean context. Having `return` on its own line will fix it.

cjprybol: Fixing this now, but rather than correct this line I'm inclined to suggest removing this case? This is how it is in both of the tables packages but it's less informative than just using the base-case report, e.g.
```julia
julia> describe(NullableArray{Any}(5))
 * All null * 
```

isn't as informative as 
```julia
julia> describe(NullableArray{Any}(5))
Summary Stats:
Length:       5
Type:         Nullable{Any}
Unique:       1
Null:         5
% Null:       100.000000
```

another option being to still throw ` * All null * ` but do so in addition to the full report (either at the beginning or the end) like so
```julia
julia> describe(NullableArray{Any}(5))
Summary Stats:
Length:       5
Type:         Nullable{Any}
Unique:       1
Null:         5
% Null:       100.000000
 * All null *
```

although that last option is more code and doesn't say anything more than `% Null:       100.000000` could tell us.

ararslan: I don't particularly mind either way, though I guess the second option there without the explicit "all null" part would be preferable for consistency.

src/utils.jl
```diff
@@ -0,0 +1,49 @@
+StatsBase.describe(X::NullableVector) = StatsBase.describe(STDOUT, X)
+
+function StatsBase.describe(io::IO, X::NullableVector{<:Real})
```
ararslan: Since this package supports 0.5 (unlike DataArrays), the `T{<:S}` syntax won't work. Not sure whether there's an `@compat` for it, but you could just do
```julia
function StatsBase.describe{T<:Real}(io::IO, X::NullableVector{T})
```

cjprybol:
Ok this is passing on 0.5 & 0.6 locally. Of course, it won't pass CI unless https://github.com/JuliaStats/StatsBase.jl/pull/247 is merged & tagged & I update the REQUIRE file. If the output as shown in the test cases is acceptable then I'll start getting the DataArrays PR ready.

test/utils.jl
```diff
@@ -0,0 +1,139 @@
...
+        @test String(take!(io)) == """
+                                   Summary Stats:
+                                   Mean:         5.500000
+                                   Minimum:      1.000000
+                                   1st Quartile: 3.250000
+                                   Median:       5.500000
+                                   3rd Quartile: 7.750000
+                                   Maximum:      10.000000
+                                   """
+        describe(io, ["s"])
```
kleinschmidt: Why test the non-nullable methods here?

kleinschmidt (commented):
LGTM except for tests which target non-nullable description.

test/utils.jl
cjprybol: Sanity checks for method ambiguity errors. Having confirmed there aren't any, the most significant thing they add is probably CI runtime. I've take them out. Thanks!

src/utils.jl
```diff
@@ -0,0 +1,53 @@
...
+    println(io, "Unique:       $(length(unique(X)))")
+    println(io, "Null:         $(nullcount)")
+    @printf(io, "%% Null:       %.6f\n", pnull)
+    return
+end
+
+function StatsBase.describe{T<:Real}(io::IO, X::AbstractVector{Nullable{T}})
+    nullcount = sum(_isnull, X)
+    pnull = 100nullcount/length(X)
+    if pnull != 100 # describe will fail if dropnull returns an empty vector
```
nalimilan, line 32: Would make sense to change this behaviour in StatsBase. The failure comes from `quantile`, where raising an error is legitimate, but for a function like `describe`, printing something about the vector being empty would be better. Can wait for after this PR, though.

src/utils.jl
```diff
@@ -0,0 +1,53 @@
...
+        describe(io, dropnull(X))
+    else
+        println(io, "Summary Stats:")
+        println(io, "Type:         $(eltype(X))")
+    end
+    println(io, "Null:         $(nullcount)")
```
nalimilan: Maybe say "Number Missing" and "% Missing" or something like that? That would be closer to the statistics terminology, and would allow hiding implementation differences between `DataArray` and `NullableArray`.

cjprybol: I like that idea! I went with "Missing" rather than "Number Missing" as we don't say "Number Unique". I guess we could say "Number ..." for both of them, but everything lines up nicer without the "Number " prefix.

ararslan: "N Missing" would also be common terminology

nalimilan: Actually I had the same reaction when reading "unique": it would be more explicit to say something to indicate that it's a count. "N" might be good enough, but "Number" is even clearer and it's not too long either.

REQUIRE
```diff
@@ -1,3 +1,4 @@
 julia 0.5
 Compat 0.19.0
 Reexport
+StatsBase
```
nalimilan: Should require the latest version of StatsBase now that it's been tagged. Otherwise, should be good to go.

REQUIRE
```diff
@@ -1,3 +1,4 @@
 julia 0.5
 Compat 0.19.0
 Reexport
+StatsBase 0.14.1
```
nalimilan: As noted elsewhere, I think this should be 0.14, right? Else Tony is going to scream at us. :-)

quinnj:
Bump, anything else to do here?

cjprybol:
Should be ready!

nalimilan:
Thanks!

PR #194, opened 2017-05-16
fix warning from overriding Base.==

shashi:
Should fix this:
```
WARNING: Method definition ==(Base.Nullable{S}, Base.Nullable{T}) in module Base at nullable.jl:244 overwritten in module NullableArrays at /home/shashi/.julia/v0.6/NullableArrays/src/operators.jl:128.
```

### DIFF
--- a/src/operators.jl
+++ b/src/operators.jl
@@ -123,6 +123,11 @@ end
 for op in (:+, :-, :*, :/, :%, :÷, :&, :|, :^, :<<, :>>, :(>>>),
            :(==), :<, :>, :<=, :>=,
            :scalarmin, :scalarmax)
+
+    if method_exists(getfield(Base, op), Tuple{Nullable, Nullable})
+        continue
+    end
+
     @eval begin
         @inline function $op{S,T}(x::Nullable{S}, y::Nullable{T})
             R = promote_op(@functorize($op), S, T)


nalimilan:
Unfortunately, that won't work. We are actually overwriting (!) the Base method so that `==` doesn't throw an error. What we should do instead is deprecate `==` in favor of `.==` in Julia 0.6, and implement the new `.==` behavior ourselves (which is the same code as the one for `==` we already have) on Julia 0.5. That's quite a bit of work.

shashi:
:cry:

shashi:
FWIW, I think
```
julia> Nullable(1) == Nullable(2)
Nullable{Bool}(false)
```
is wrong and would prefer the Base behavior. (Really hate the `got Nullable{Bool} where Bool was expected` errors in if statements written for generic values. This is the only type that messes with the return type of `==`). I can accept that this is the behavior of `.==` though.

> and implement the new .== behavior ourselves (which is the same code as the one for == we already have) on Julia 0.5. 

we already have where? and also did you mean 0.6?

nalimilan:
We have it in this package, since that's what this issue is about!

I meant Julia 0.5: people cannot use `.==` as a replacement for `==` there. But we could also drop support for it.

PR #202, opened 2017-08-15
Ensure that `copy` and `getindex` maintain `parent` field

### LINKED ISSUE #199: `copy` and `deepcopy` do not copy the `parent` field
I noticed this issue when sorting a `DataTable` that contained a column of type `NullableArray{WeakRefString{UInt8}}`. Ultimately `sort` calls `copy`, and after the smoke cleared I was left with `WeakRefString`s that would occasionally become gibberish because garbage collection happened and there were no longer any open references to the block of data used by the `WeakRefString`s.

I'm putting together a PR that I hope will fix the issue with `copy`, but the issues with `deepcopy` (if you'll excuse the expression) run deeper.

As the `parent` field is really just a reference, we don't *want* a deep copy of it. Currently, if you `deepcopy`, you just get an empty `parent`, which won't prevent garbage collection of any unsafe references (like `WeakRefString`s). If we wanted `deepcopy` to work in this case, it might end up looking something like this:

```julia
function Base.deepcopy{T, N}(X::NullableArray{T, N})
    return NullableArray{T, N}(deepcopy(X.values), deepcopy(X.isnull), X.parent)
end
```

In addition to not *really* being a deep copy, this has the unfortunate consequence that if you're copying a `NullableArray` that doesn't care about its `parent` field (which seems to be most of them), you end up with an original and a copy that share a reference to a single `parent` (which is just going to be an empty `Vector{UInt8}`). It's maybe not a really big deal, but it certainly *seems wrong*.

So my plan is to attempt to fix `copy`, and to leave `deepcopy` alone. I don't know if the long-term plan for `NullableArrays` is to keep `parent` around (my understanding is that it is not), but the fact that it isn't being copied is causing problems for us at the moment.

spurll:
Closes #199

### DIFF
--- a/test/indexing.jl
+++ b/test/indexing.jl
@@ -57,6 +57,12 @@ module TestIndexing
         end
     end
 
+    # ensure that the NullableArray returned by getindex maintains its parent field
+    str = "something"
+    x = NullableArray([pointer(str), pointer(str, 5)], [false, false], Vector{UInt8}(str))
+    @test isequal(x[1:2].parent, x.parent)
+    @test isequal(x[:].parent, x.parent)
+
     # getindex with AbstractVectors
     rg = 2:9
     v = X[rg]
--- a/test/indexing.jl
+++ b/test/indexing.jl
@@ -60,8 +60,8 @@ module TestIndexing
     # ensure that the NullableArray returned by getindex maintains its parent field
     str = "something"
     x = NullableArray([pointer(str), pointer(str, 5)], [false, false], Vector{UInt8}(str))
-    @test isequal(x[1:2].parent, x.parent)
-    @test isequal(x[:].parent, x.parent)
+    @test x[1:2].parent === x.parent
+    @test x[:].parent === x.parent
 
     # getindex with AbstractVectors
     rg = 2:9
--- a/src/constructors.jl
+++ b/src/constructors.jl
@@ -5,8 +5,9 @@
 # that of the inner constructor provided in typedefs.jl: constructs a NullableArray
 # from an AbstractArray of values and an AbstractArray{Bool} mask.
 function NullableArray{T, N}(A::AbstractArray{T, N},
-                             m::AbstractArray{Bool, N}) # -> NullableArray{T, N}
-    return NullableArray{T, N}(A, m)
+                             m::AbstractArray{Bool, N},
+                             parent::Vector{UInt8}=Vector{UInt8}()) # -> NullableArray{T, N}
+    return NullableArray{T, N}(A, m, parent)
 end
 
 # TODO: Uncomment this doc entry when Base Julia can parse it correctly.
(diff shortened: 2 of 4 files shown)


spurll:
Hm, I think that Codecov report is wrong, since I didn't touch most of those impacted files.

However, I realise that I'll need to make another commit to solve an outstanding issue related to sorting `DataTables`, so I'm going to mark this WiP.

ararslan:
> Hm, I think that Codecov report is wrong

Yeah, coverage is weird.

> However, I realise that I'll need to make another commit to solve an outstanding issue related to sorting DataTables

Is that directly related to this change?

spurll:
The problem I'm currently trying to solve is the fact that the new `NullableArray` created by `getindex` doesn't maintain the parent reference either:

```julia
julia> str = "something"
"something"

julia> arr = [WeakRefString(pointer(str, 1), 4, 1), WeakRefString(pointer(str, 5), 5, 5)]
2-element Array{WeakRefString{UInt8},1}:
 "some" 
 "thing"

julia> na = NullableArray{WeakRefString{UInt8}, 1}(arr, [false, false], Vector{UInt8}(str))
2-element NullableArrays.NullableArray{WeakRefString{UInt8},1}:
 "some" 
 "thing"

julia> na.parent
9-element Array{UInt8,1}:
 0x73
 0x6f
 0x6d
 0x65
 0x74
 0x68
 0x69
 0x6e
 0x67

julia> na[:].parent
0-element Array{UInt8,1}
```

It's certainly related, though maybe it warrants a separate PR.

spurll:
This should now be good to go. Though it's possible that the change I made to `similar` will be contentious, it is necessary to maintain the parent reference without massive changes to the codebase.

nalimilan (approved):
Thanks, this makes sense, though that `parent` field is really an ugly hack... I'd like a check by @quinnj, who introduced this feature.

test/indexing.jl
```diff
@@ -57,6 +57,12 @@ module TestIndexing
         end
     end
 
+    # ensure that the NullableArray returned by getindex maintains its parent field
+    str = "something"
+    x = NullableArray([pointer(str), pointer(str, 5)], [false, false], Vector{UInt8}(str))
+    @test isequal(x[1:2].parent, x.parent)
```
nalimilan: Use `===` Same below..

nalimilan (changes requested):
Woops, I had missed some details.

test/indexing.jl
spurll: Thanks. I've made that change, and the corresponding change to test/primitives.jl as well.

quinnj (approved):
This looks fine. I have new code in WeakRefStrings that introduces a WeakRefStringArray so that we don't need this kind of `parent` field anymore. I'm going to try and do new releases for all of DataStreams this week that uses it.

PR #203, opened 2017-08-15
Allow 0.7 failures, test on 0.6, fix importall deprecation

ararslan:


src/operators.jl
```diff
@@ -1,8 +1,8 @@
 ## Lifted operators
 
-importall Base.Operators
-import Base: promote_op, abs, abs2, sqrt, cbrt, scalarmin, scalarmax, isless
-using Compat: @compat
+import Base: !, %, &, *, +, -, /, //, <, <<, <<<, ==, ===, >, >>, >>>, \, ^, getindex, |, ~,
```
ararslan: I was mostly just guessing as to which operators need to be imported. Some of these are probably not necessary.

rofinn: Did a `setdiff` and it looks like this covers all the operators being overloaded and the only extras are `Set(Symbol[:\, ://, :(===)])` which aren't unreasonable.

.travis.yml
```diff
@@ -1,10 +1,18 @@
...
-    - julia -e 'Pkg.test("NullableArrays", coverage=true)'
+  - 0.5
+  - 0.6
+  - nightly
+matrix:
+  allow_failures:
+    - julia: nightly
+notifications:
+  email: false
+#script:
```
rofinn, line 13: Forgot to uncomment?

rofinn (changes requested):
Looks good once you've uncommented the `script` in .travis.yml.

.travis.yml
ararslan: Commenting it allows the default script to run, which generally shouldn't be overridden unless there's a good reason to do so.

rofinn: Oh, I didn't realize there was a default script.

src/operators.jl
```diff
@@ -89,58 +76,42 @@ end
...
-            null_safe_op{S<:SafeUnsigned,
-                         T<:SafeUnsigned}(::typeof($op), ::Type{S}, ::Type{T}) = true
-        end
+for op in (:+, :-, :*, :/, :&, :|, :<<, :>>, :(>>>),
+           :(==), :<, :>, :<=, :>=,
+           :scalarmin, :scalarmax,
+           :isless)
+    @eval begin
+        # to fix ambiguities
+        null_safe_op{S<:SafeFloats,
```
rofinn, line 85: Seems more consistent to do `::typeof(Base.$op)` here as well rather than importing `scalarmin` and `scalarmax` above?

.travis.yml
ararslan: Yep. Useful to have in case the default needs to change for whatever reason.

src/operators.jl
ararslan: Yes, I had that but deleted it in favor of the scalarmin/scalarmax stuff. Idk why. 🙃

ararslan: Actually I think I prefer it this way, since it shows that the explicit prefix implies extension and no prefix implies using the function. But if you feel strongly I can change it.

rofinn: No, that's fine I guess.

rofinn (approved):

